PR #3, opened 2022-06-22
Restructure modules

ajhelsby:


.github/workflows/build.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Create and activate Virtualenv
+        run: |
+          pip install virtualenv
+          [ ! -d ".venv" ] && virtualenv .venv
+          . .venv/bin/activate
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements-dev.txt
```
matthewelwell, line 38: I think you'll need to install requirements.txt here too? You can do that in a single command 

`pip install -r requirements.txt requirements-dev.txt`

.github/workflows/build.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Run isort formatter check
+        run: |
+          isort .
+
+      - name: Test with pytest
+        run: |
+          pytest
+
+  package:
```
matthewelwell, line 56: I think let's remove this job for now as it's not really doing anything?

pyproject.toml
```diff
@@ -1,3 +1,30 @@
...
+[build-system]
+requires      = ["setuptools>=61.0.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "python_open_feature_sdk"
+version = "0.0.1"
+description = "Standardizing Feature Flagging for Everyone"
+readme = "README.md"
+authors = [{ name = "Real Python", email = "info@realpython.com" }]
```
matthewelwell, line 11: This will need updating - looks like this was probably copy / pasted from somewhere?

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+  push:
+    branches:
+      - '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container: [ "python:3.8", "python:3.9", "python:3.10" ]
```
matthewelwell, line 18: Interesting... I didn't realise you could do this. Is there any benefit to doing it this way rather than using the default ubuntu image and setting up the python version using the relevant action. 

Like this: https://github.com/Flagsmith/flagsmith/blob/main/.github/workflows/api-pull-request.yml#L41

pyproject.toml
```diff
@@ -1,3 +1,30 @@
+# pyproject.toml
+[build-system]
+requires      = ["setuptools>=61.0.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "python_open_feature_sdk"
```
matthewelwell, line 7: We may want to change this - I'm not sure if there are any conventions we need to follow. But it's not important for this PR.

matthewelwell (approved):
Approved with a couple of minor comments.

pyproject.toml
ajhelsby: Easy enough to change at another point

matthewelwell: Agreed.

matthewelwell:
@toddbaert @justinabrahms would you like to review this before we merge? I guess generally, would you prefer us to work autonomously on this or would you like to review each PR before it's merged?

matthewelwell:
Understood, thanks @justinabrahms.

PR #4, opened 2022-06-24
add flags type using factory design pattern

yousef-soliman:
it's better to use the factory design pattern for the flags type 
I made a POC to start discussing the idea of using the factory design pattern

### DIFF
--- a/src/flag/abstact_flag.py
+++ b/src/flag/abstact_flag.py
@@ -1,0 +1,17 @@
+import typing
+
+
+class AbstractFlag:
+    def __init__(
+        self,
+        key: str,
+        evaluation_context: typing.Any,
+        flag_evaluation_context typing.Any,
+    ) -> None:
+        self.key = key
+        self.evaluation_context = evaluation_context
+        self.flag_evaluation_context = flag_evaluation_context
+
+    @property
+    def value(self):
+        raise NotImplementedError()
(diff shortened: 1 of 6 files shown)


toddbaert:
Thanks for your contribution!

These are all new files, I don't see them being used anywhere. The OpenFeature SDK is for evaluating flags, not defining them, generally, so I'm not sure such classes are needed, but maybe!

I'll leave that up to the people who've contributed to this SDK so far.

justinabrahms:
tbh, this feels pretty java-y and less pythonic. I know that's a bit wishy-washy of a feedback mechanism.

Can you say more about why you want this change rather than "it's better"?

yousef-soliman:
@justinabrahms @toddbaert  Thanks for feedback
These are my thoughts to use the Factory design pattern

- it reduces code duplication 
you don't need to write these ifs again when you need to determine the flag type
```python
if flag_type is FlagType.BOOLEAN:
                return self.provider.get_boolean_details(*args)

            elif flag_type is FlagType.NUMBER:
                return self.provider.get_number_details(*args)

            elif flag_type is FlagType.OBJECT:
                return self.provider.get_object_details(*args)
            # Fallback case is a string object
            elif flag_type is FlagType.STRING:
                return self.provider.get_string_details(*args)
            else:

```
just use FlagClass and call your function 
```python
any_flag().get_details()
```

- extendable 
when you need to add a new type of flag just define the flag and basic functions
no need to go through every IFs to change and elif  

- abstract  layer of type 
```python
any_flag().get_value()
```

PR #5, opened 2022-07-01
evaluation-context: Add evaluation context to open feature

ajhelsby:
An initial implementation of the evaluation context spec of Open Feature seen here https://github.com/open-feature/spec/blob/main/specification/evaluation-context.md

### DIFF
--- a/open_feature/evaluation_context/evaluation_context.py
+++ b/open_feature/evaluation_context/evaluation_context.py
@@ -1,0 +1,17 @@
+class EvaluationContext:
+    def __init__(self, targeting_key: str = None, attributes: dict = None):
+        self.targeting_key = targeting_key
+        self.attributes = attributes
+
+    def merge(
+        self, ctx1: "EvaluationContext", ctx2: "EvaluationContext"
+    ) -> "EvaluationContext":
+        self.attributes = ctx1.attributes.update(ctx2.attributes)
+
+        if ctx1.targeting_key is not None:
+            self.targeting_key = ctx1.targeting_key
+
+        if ctx2.targeting_key is not None:
+            self.targeting_key = ctx2.targeting_key
+
+        return self
--- a/open_feature/evaluation_context/evaluation_context.py
+++ b/open_feature/evaluation_context/evaluation_context.py
@@ -1,5 +1,5 @@
 class EvaluationContext:
-    def __init__(self, targeting_key: str = None, attributes: dict = None):
+    def __init__(self, targeting_key: str = None, attributes: dict = {}):
         self.targeting_key = targeting_key
         self.attributes = attributes
 
--- a/open_feature/evaluation_context/evaluation_context.py
+++ b/open_feature/evaluation_context/evaluation_context.py
@@ -1,7 +1,7 @@
 class EvaluationContext:
-    def __init__(self, targeting_key: str = None, attributes: dict = {}):
+    def __init__(self, targeting_key: str = None, attributes: dict = None):
         self.targeting_key = targeting_key
-        self.attributes = attributes
+        self.attributes = attributes or {}
 
     def merge(
         self, ctx1: "EvaluationContext", ctx2: "EvaluationContext"
--- a/open_feature/evaluation_context/evaluation_context.py
+++ b/open_feature/evaluation_context/evaluation_context.py
@@ -4,7 +4,7 @@ class EvaluationContext:
         self.attributes = attributes or {}
 
     def merge(self, ctx2: "EvaluationContext") -> "EvaluationContext":
-        self.attributes = self.attributes | ctx2.attributes
+        self.attributes = {**self.attributes, **ctx2.attributes}
         self.targeting_key = ctx2.targeting_key or self.targeting_key
 
         return self
--- a/open_feature/evaluation_context/evaluation_context.py
+++ b/open_feature/evaluation_context/evaluation_context.py
@@ -3,10 +3,8 @@ class EvaluationContext:
         self.targeting_key = targeting_key
         self.attributes = attributes or {}
 
-    def merge(
-        self, ctx1: "EvaluationContext", ctx2: "EvaluationContext"
-    ) -> "EvaluationContext":
-        self.attributes = ctx1.attributes.update(ctx2.attributes)
-        self.targeting_key = ctx2.targeting_key or ctx1.targeting_key
+    def merge(self, ctx2: "EvaluationContext") -> "EvaluationContext":
+        self.attributes = self.attributes | ctx2.attributes
+        self.targeting_key = ctx2.targeting_key or self.targeting_key
 
         return self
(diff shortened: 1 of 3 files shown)


open_feature/evaluation_context/evaluation_context.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, targeting_key: str = None, attributes: dict = {}):
+        self.targeting_key = targeting_key
+        self.attributes = attributes
+
+    def merge(
+        self, ctx1: "EvaluationContext", ctx2: "EvaluationContext"
+    ) -> "EvaluationContext":
+        self.attributes = ctx1.attributes.update(ctx2.attributes)
+
+        if ctx1.targeting_key is not None:
```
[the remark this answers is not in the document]
matthewelwell: This does assume, however, that `""` is not a valid targeting_key (which I think is a valid assumption?). 

There might be some other edge cases where this behaves differently to your code?

open_feature/open_feature_client.py
```diff
@@ -149,34 +149,54 @@ def evaluate_flag_details(
...
         default_value: typing.Any,
-        evaluation_context: typing.Any = None,
+        evaluation_context: EvaluationContext = None,
         flag_evaluation_options: typing.Any = None,
     ) -> FlagEvaluationDetails:
+        if evaluation_context is None:
+            evaluation_context = EvaluationContext()
         try:
-            args = (
+            # The merge below will eventually pull in Evaluation contexts from Hooks
```
matthewelwell, line 158: Do we need this code here yet? Should we wait until we have the hooks before adding these lines. It's a bit confusing at the moment?

open_feature/open_feature_client.py
```diff
@@ -149,34 +149,54 @@ def evaluate_flag_details(
...
-            elif flag_type is FlagType.OBJECT:
-                return self.provider.get_object_details(*args)
-            # Fallback case is a string object
-            elif flag_type is FlagType.STRING:
-                return self.provider.get_string_details(*args)
-            else:
-                raise GeneralError(error_message="Unknown flag type")
-
-        except OpenFeatureError:
+        except OpenFeatureError as ofe:
```
matthewelwell, line 168: I think we will have to catch all exceptions here by the way? Something that Justin / Todd mentioned in the previous review which I think we missed here. One for a separate PR though.

open_feature/open_feature_client.py
```diff
@@ -149,34 +149,54 @@ def evaluate_flag_details(
...
             return FlagEvaluationDetails(
                 key=key,
                 value=default_value,
                 reason=Reason.ERROR,
-                error_code=ErrorCode.GENERAL,
+                error_code=ofe.error_message,
             )
+
+    def create_provider_evaluation(
```
matthewelwell, line 176: Seems like this has been ripped from the java client `createProviderEvaluation` method but I don't quite understand the logic based on the name of the method here? Perhaps we should add some docstring to this function.

open_feature/open_feature_client.py
ajhelsby: The hook code to replace this comment is ready to go as well, I just need this one in before I can add it

open_feature/open_feature_client.py
ajhelsby: That's fine, we just need to handle the error message though as we can't guarantee there is a message in a generic exception, which we use for the error code.

open_feature/evaluation_context/evaluation_context.py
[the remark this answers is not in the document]
ajhelsby: For now I think assuming `""` is not a valid target is fine but we can revisit it.

open_feature/evaluation_context/evaluation_context.py
```diff
@@ -0,0 +1,12 @@
+class EvaluationContext:
+    def __init__(self, targeting_key: str = None, attributes: dict = None):
+        self.targeting_key = targeting_key
+        self.attributes = attributes or {}
+
+    def merge(
+        self, ctx1: "EvaluationContext", ctx2: "EvaluationContext"
```
matthewelwell, line 7: Isn't `self` ctx1 in this case? Or should this be a static method? It appears that here, we're working with 3 different EvaluationContext objects which seems wrong?

open_feature/open_feature_client.py
```diff
@@ -149,34 +149,65 @@ def evaluate_flag_details(
...
+        evaluation_context: EvaluationContext = None,
         flag_evaluation_options: typing.Any = None,
     ) -> FlagEvaluationDetails:
+        if evaluation_context is None:
+            evaluation_context = EvaluationContext()
         try:
-            args = (
+            # The merge below will eventually pull in Evaluation contexts from Hooks
+            invocation_context = EvaluationContext()
+            invocation_context.merge(ctx1=EvaluationContext(), ctx2=evaluation_context)
```
matthewelwell, line 160: Related to the above - I'm not sure why we're creating 2 new EvaluationContext objects?

open_feature/evaluation_context/evaluation_context.py
```diff
@@ -0,0 +1,12 @@
+class EvaluationContext:
+    def __init__(self, targeting_key: str = None, attributes: dict = None):
+        self.targeting_key = targeting_key
+        self.attributes = attributes or {}
+
+    def merge(
```
matthewelwell, line 6: Can we write some unit tests for this?

open_feature/open_feature_client.py
```diff
@@ -149,34 +149,65 @@ def evaluate_flag_details(
...
+        :return: a FlagEvaluationDetails object with the fully evaluated flag from a
+        provider
+        """
+        args = (
+            key,
+            default_value,
+            evaluation_context,
+            flag_evaluation_options,
+        )
+        if flag_type is FlagType.BOOLEAN:
```
matthewelwell, line 201: Perhaps a slightly more pythonic (and succinct) way of writing this would be: 

```python
get_details_callable = {
    FlagType.BOOLEAN: self.provider.get_boolean_details,
    FlagType.NUMBER: self.provider.get_number_details,
    ...
}.get(flag_type)
if not get_details_callable:
    raise GeneralError(...)
return get_details_callable(*args)
```

open_feature/evaluation_context/evaluation_context.py
ajhelsby: it should be static

ajhelsby: Yes agreed, self is the first context, I've made that change now

open_feature/open_feature_client.py
ajhelsby: Yes it' unnecessary, I'll make the change

matthewelwell (approved):

open_feature/open_feature_client.py
matthewelwell: Seems like these changes were dropped entirely, are they going to be in a separate PR?

PR #6, opened 2022-08-04
hooks: Implementation of hooks for python OpenFeature

ajhelsby:
Signed-off-by: Andrew Helsby <ajhelsby@hotmail.com>

### DIFF
--- a/open_feature/hooks/hook.py
+++ b/open_feature/hooks/hook.py
@@ -1,0 +1,62 @@
+from abc import abstractmethod
+
+from open_feature.flag_evaluation.flag_evaluation_details import FlagEvaluationDetails
+from open_feature.flag_evaluation.flag_type import FlagType
+
+
+class Hook:
+    @abstractmethod
+    def before(self, ctx, hints: dict):
+        """
+        Runs before flag is resolved.
+
+        :param ctx: Information about the particular flag evaluation
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        :return: An optional EvaluationContext. If returned, it will
+        be merged with the EvaluationContext instances from other
+        hooks, the client and API.
+        """
+        pass
+
+    @abstractmethod
+    def after(self, ctx, details: FlagEvaluationDetails, hints: dict):
+        """
+        Runs after a flag is resolved.
+
+        :param ctx: Information about the particular flag evaluation
+        :param details: Information about how the flag was resolved,
+        including any resolved values.
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        """
+        pass
+
+    @abstractmethod
+    def error(self, ctx, exception: Exception, hints: dict):
+        """
+        Run when evaluation encounters an error. This will always run.
+        Errors thrown will be swallowed.
+
+        :param ctx: Information about the particular flag evaluation
+        :param exception:
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        """
+        pass
+
+    @abstractmethod
+    def finally_after(self, ctx, hints: dict):
+        """
+        Run after flag evaluation, including any error processing.
+        This will always run. Errors will be swallowed.
+
+        :param ctx: Information about the particular flag evaluation
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        """
+        pass
+
+    @abstractmethod
+    def supports_flag_value_type(self, flag_type: FlagType) -> bool:
+        return True
--- a/open_feature/hooks/hook.py
+++ b/open_feature/hooks/hook.py
@@ -1,16 +1,18 @@
 from abc import abstractmethod
 
+from open_feature.evaluation_context.evaluation_context import EvaluationContext
 from open_feature.flag_evaluation.flag_evaluation_details import FlagEvaluationDetails
 from open_feature.flag_evaluation.flag_type import FlagType
+from open_feature.hooks.hook_context import HookContext
 
 
 class Hook:
     @abstractmethod
-    def before(self, ctx, hints: dict):
+    def before(self, hook_context: HookContext, hints: dict) -> EvaluationContext:
         """
         Runs before flag is resolved.
 
-        :param ctx: Information about the particular flag evaluation
+        :param hook_context: Information about the particular flag evaluation
         :param hints: An immutable mapping of data for users to
         communicate to the hooks.
         :return: An optional EvaluationContext. If returned, it will
@@ -20,11 +22,13 @@ class Hook:
         pass
 
     @abstractmethod
-    def after(self, ctx, details: FlagEvaluationDetails, hints: dict):
+    def after(
+        self, hook_context: HookContext, details: FlagEvaluationDetails, hints: dict
+    ):
         """
         Runs after a flag is resolved.
 
-        :param ctx: Information about the particular flag evaluation
+        :param hook_context: Information about the particular flag evaluation
         :param details: Information about how the flag was resolved,
         including any resolved values.
         :param hints: An immutable mapping of data for users to
@@ -33,12 +37,12 @@ class Hook:
         pass
 
     @abstractmethod
-    def error(self, ctx, exception: Exception, hints: dict):
+    def error(self, hook_context: HookContext, exception: Exception, hints: dict):
         """
         Run when evaluation encounters an error. This will always run.
         Errors thrown will be swallowed.
 
-        :param ctx: Information about the particular flag evaluation
+        :param hook_context: Information about the particular flag evaluation
         :param exception:
         :param hints: An immutable mapping of data for users to
         communicate to the hooks.
@@ -46,12 +50,12 @@ class Hook:
         pass
 
     @abstractmethod
-    def finally_after(self, ctx, hints: dict):
+    def finally_after(self, hook_context: HookContext, hints: dict):
         """
         Run after flag evaluation, including any error processing.
         This will always run. Errors will be swallowed.
 
-        :param ctx: Information about the particular flag evaluation
+        :param hook_context: Information about the particular flag evaluation
         :param hints: An immutable mapping of data for users to
         communicate to the hooks.
         """
--- a/open_feature/hooks/hook.py
+++ b/open_feature/hooks/hook.py
@@ -40,7 +40,7 @@ class Hook:
         Run when evaluation encounters an error. Errors thrown will be swallowed.
 
         :param hook_context: Information about the particular flag evaluation
-        :param exception: Ability to throw a specific exception if required
+        :param exception: The exception that was thrown
         :param hints: A mapping of data for users to communicate to the hooks.
         """
         pass
--- a/open_feature/hooks/hook.py
+++ b/open_feature/hooks/hook.py
@@ -15,9 +15,8 @@ class Hook:
         :param hook_context: Information about the particular flag evaluation
         :param hints: An immutable mapping of data for users to
         communicate to the hooks.
-        :return: An EvaluationContext. If returned, it will
-        be merged with the EvaluationContext instances from other
-        hooks, the client and API.
+        :return: An EvaluationContext. It will be merged with the
+        EvaluationContext instances from other hooks, the client and API.
         """
         pass
 
--- a/open_feature/hooks/hook.py
+++ b/open_feature/hooks/hook.py
@@ -15,7 +15,7 @@ class Hook:
         :param hook_context: Information about the particular flag evaluation
         :param hints: An immutable mapping of data for users to
         communicate to the hooks.
-        :return: An optional EvaluationContext. If returned, it will
+        :return: An EvaluationContext. If returned, it will
         be merged with the EvaluationContext instances from other
         hooks, the client and API.
         """
@@ -31,21 +31,18 @@ class Hook:
         :param hook_context: Information about the particular flag evaluation
         :param details: Information about how the flag was resolved,
         including any resolved values.
-        :param hints: An immutable mapping of data for users to
-        communicate to the hooks.
+        :param hints: A mapping of data for users to communicate to the hooks.
         """
         pass
 
     @abstractmethod
     def error(self, hook_context: HookContext, exception: Exception, hints: dict):
         """
-        Run when evaluation encounters an error. This will always run.
-        Errors thrown will be swallowed.
+        Run when evaluation encounters an error. Errors thrown will be swallowed.
 
         :param hook_context: Information about the particular flag evaluation
-        :param exception:
-        :param hints: An immutable mapping of data for users to
-        communicate to the hooks.
+        :param exception: Ability to throw a specific exception if required
+        :param hints: A mapping of data for users to communicate to the hooks.
         """
         pass
 
@@ -56,11 +53,17 @@ class Hook:
         This will always run. Errors will be swallowed.
 
         :param hook_context: Information about the particular flag evaluation
-        :param hints: An immutable mapping of data for users to
-        communicate to the hooks.
+        :param hints: A mapping of data for users to communicate to the hooks.
         """
         pass
 
     @abstractmethod
     def supports_flag_value_type(self, flag_type: FlagType) -> bool:
+        """
+        Check to see if the hook supports the particular flag type.
+
+        :param flag_type: particular type of the flag
+        :return: a boolean containing whether the flag type is supported (True)
+        or not (False)
+        """
         return True
--- a/open_feature/hooks/hook_context.py
+++ b/open_feature/hooks/hook_context.py
@@ -1,0 +1,19 @@
+from open_feature.flag_evaluation.flag_type import FlagType
+
+
+class HookContext:
+    def __init__(  # noqa: CFQ002
+        self,
+        flag_key: str,
+        flag_type: FlagType,
+        default_value,
+        ctx,
+        client_metadata=None,
+        provider_metadata=None,
+    ):
+        self.flag_key = flag_key
+        self.flag_type = flag_type
+        self.default_value = default_value
+        self.ctx = ctx
+        self.client_metadata = client_metadata
+        self.provider_metadata = provider_metadata
--- a/open_feature/hooks/hook_context.py
+++ b/open_feature/hooks/hook_context.py
@@ -1,3 +1,4 @@
+from open_feature.evaluation_context.evaluation_context import EvaluationContext
 from open_feature.flag_evaluation.flag_type import FlagType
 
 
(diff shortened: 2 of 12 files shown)


open_feature/hooks/hook.py
```diff
@@ -0,0 +1,66 @@
...
+class Hook:
+    @abstractmethod
+    def before(self, hook_context: HookContext, hints: dict) -> EvaluationContext:
+        """
+        Runs before flag is resolved.
+
+        :param hook_context: Information about the particular flag evaluation
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        :return: An optional EvaluationContext. If returned, it will
```
matthewelwell, line 18: This seems to indicate that the typehint above should be `typing.Optional[EvaluationContext]`

open_feature/hooks/hook.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Hook:
+    @abstractmethod
+    def before(self, hook_context: HookContext, hints: dict) -> EvaluationContext:
+        """
+        Runs before flag is resolved.
+
+        :param hook_context: Information about the particular flag evaluation
+        :param hints: An immutable mapping of data for users to
```
matthewelwell, line 16: Not sure how / if / why we intend to ensure that the hints dict here is immutable?

open_feature/hooks/hook.py
```diff
@@ -0,0 +1,66 @@
...
+        including any resolved values.
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        """
+        pass
+
+    @abstractmethod
+    def error(self, hook_context: HookContext, exception: Exception, hints: dict):
+        """
+        Run when evaluation encounters an error. This will always run.
```
matthewelwell, line 42: `'This will always run'` seems misleading here? IS this supposed to only be on `finally_after` ?

open_feature/hooks/hook.py
```diff
@@ -0,0 +1,66 @@
...
+        Errors thrown will be swallowed.
+
+        :param hook_context: Information about the particular flag evaluation
+        :param exception:
```
matthewelwell, line 46: Missing a description of the exception argument

open_feature/hooks/hook.py
```diff
@@ -0,0 +1,66 @@
...
+        This will always run. Errors will be swallowed.
+
+        :param hook_context: Information about the particular flag evaluation
+        :param hints: An immutable mapping of data for users to
+        communicate to the hooks.
+        """
+        pass
+
+    @abstractmethod
+    def supports_flag_value_type(self, flag_type: FlagType) -> bool:
```
matthewelwell, line 65: Should we add a docstring here?

open_feature/hooks/hook_context.py
```diff
@@ -0,0 +1,20 @@
+from open_feature.evaluation_context.evaluation_context import EvaluationContext
+from open_feature.flag_evaluation.flag_type import FlagType
+
+
+class HookContext:
+    def __init__(  # noqa: CFQ002
+        self,
+        flag_key: str,
+        flag_type: FlagType,
+        default_value,
```
matthewelwell, line 10: Missing typehint? `typing.Any` would suffice if it can be any type.

open_feature/hooks/hook_context.py
```diff
@@ -0,0 +1,20 @@
...
+        evaluation_context: EvaluationContext,
+        client_metadata=None,
```
matthewelwell, line 12: Missing typehints?

open_feature/hooks/hook_context.py
```diff
@@ -0,0 +1,20 @@
+from open_feature.evaluation_context.evaluation_context import EvaluationContext
+from open_feature.flag_evaluation.flag_type import FlagType
+
+
+class HookContext:
```
matthewelwell, line 5: Could this be simplified to a python `dataclass` ?

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+    exception: Exception,
+    hooks: typing.List[Hook],
+    hints: dict,
+):
+    kwargs = {"ctx": hook_context, "exception": exception, "hints": hints}
+    execute_hooks(flag_type=flag_type, hooks=hooks, hook_method="error", **kwargs)
+
+
+def after_all_hooks(
+    flag_type: FlagType, hook_context: HookContext, hooks: list, hints: dict
```
matthewelwell, line 24:
```suggestion
    flag_type: FlagType, hook_context: HookContext, hooks: typing.List[Hook], hints: dict
```

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+
+def before_hooks(
+    flag_type: FlagType,
+    hook_context: HookContext,
+    hooks: typing.List[Hook],
+    hints: dict,
+) -> EvaluationContext:
+    kwargs = {"ctx": hook_context, "hints": hints}
+    executed_hooks = execute_hooks(
+        flag_type=flag_type, hooks=hooks, hook_method="before", **kwargs
```
matthewelwell, line 53: Might be worth creating an enum or constants for `"before"`, `"after"`, etc. just to avoid any programming errors.

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    filtered_hooks = list(filter(lambda hook: hook is None, executed_hooks))
```
matthewelwell, line 55: Shouldn't this be `lambda hook: hook is not None` ?

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if filtered_hooks:
+        return reduce(lambda a, b: a.merge(b), filtered_hooks)
+    else:
```
matthewelwell, line 59: This is really just code style preferences but it's slightly more pythonic to do:

```python
if filtered_hooks:
    return ...

return EvaluationContext()
```

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+        return EvaluationContext()
+
+
+def execute_hooks(
+    flag_type: FlagType, hooks: typing.List[Hook], hook_method: str, **kwargs
+) -> list:
+    if hooks:
+        filtered_hooks = list(
+            filter(
+                lambda hook: hook.supports_flag_value_type(flag_type=flag_type), hooks
```
matthewelwell, line 69: It's not immediately clear to me why we only execute hooks that support flag value type?

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    filtered_hooks = list(filter(lambda hook: hook is None, executed_hooks))
+
+    if filtered_hooks:
+        return reduce(lambda a, b: a.merge(b), filtered_hooks)
+    else:
+        return EvaluationContext()
+
+
+def execute_hooks(
```
matthewelwell, line 63: I think this is probably only called by the functions defined above so could this be marked as a private function? `_execute_hooks()` (question also applies to the 2 functions below)

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+        filtered_hooks = list(
+            filter(
+                lambda hook: hook.supports_flag_value_type(flag_type=flag_type), hooks
+            )
+        )
+        return [execute_checked(hook, hook_method, **kwargs) for hook in filtered_hooks]
+    return []
+
+
+def execute_hooks_unchecked(
```
matthewelwell, line 76: Is it worth adding some documentation somewhere that explains why we sometimes run the hooks unchecked but other times, checked? It's not immediately clear to me.

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+    flag_type: FlagType, hooks, hook_method: str, **kwargs
+) -> list:
+    if hooks:
+        filtered_hooks = list(
+            filter(
+                lambda hook: hook.supports_flag_value_type(flag_type=flag_type), hooks
+            )
+        )
+        return [getattr(hook, hook_method)(**kwargs) for hook in filtered_hooks]
```
matthewelwell, line 85: We should add a default return to this function.

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+            filter(
+                lambda hook: hook.supports_flag_value_type(flag_type=flag_type), hooks
+            )
+        )
+        return [execute_checked(hook, hook_method, **kwargs) for hook in filtered_hooks]
+    return []
+
+
+def execute_hooks_unchecked(
+    flag_type: FlagType, hooks, hook_method: str, **kwargs
```
matthewelwell, line 77:
```suggestion
    flag_type: FlagType, hooks: typing.List[Hook], hook_method: str, **kwargs
```

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,92 @@
...
+    if hooks:
+        filtered_hooks = list(
+            filter(
+                lambda hook: hook.supports_flag_value_type(flag_type=flag_type), hooks
+            )
+        )
+        return [getattr(hook, hook_method)(**kwargs) for hook in filtered_hooks]
+
+
+def execute_checked(hook: Hook, hook_method: str, **kwargs):
```
matthewelwell, line 88: Should this be `execute_hook_checked` just for consistency. I was a bit confused by the `execute_checked` / `execute_hooks_checked` functions.

open_feature/open_feature_api.py
```diff
@@ -1,10 +1,12 @@
 import typing
 
+from open_feature.evaluation_context.evaluation_context import EvaluationContext
 from open_feature.exception.exceptions import GeneralError
 from open_feature.open_feature_client import OpenFeatureClient
 from open_feature.provider.provider import AbstractProvider
 
 _provider = None
+_evaluation_context = EvaluationContext()
```
matthewelwell, line 9: Should this live in the `open_feature_evaluation_context` module?

open_feature/open_feature_client.py
```diff
@@ -14,8 +25,8 @@ def __init__(
         self,
         name: str,
         version: str,
-        context=None,
-        hooks: list = None,
+        context: EvaluationContext = EvaluationContext(),
```
matthewelwell, line 28: I wonder if the `evaluation_context` module should export a constant like `EMPTY_EVALUATION_CONTEXT` to handle this? Although, actually I think that might not work. My concern is that I have a feeling python will (loosely) complain about this as we're setting a mutable argument as the default here. 

We may need to do: 

```python
def __init__(..., context: EvaluationContext = None, ...):
    ...
    self.evaluation_context = context or EvaluationContext()
```

open_feature/open_feature_client.py
```diff
@@ -152,27 +166,65 @@ def evaluate_flag_details(
...
+            client_metadata=None,
+            provider_metadata=None,
+        )
+        merged_hooks = []
+
         try:
-            # The merge below will eventually pull in Evaluation contexts from Hooks
-            invocation_context = EvaluationContext()
+            invocation_context = before_hooks(
+                flag_type, hook_context, merged_hooks, None
```
matthewelwell, line 196: How come we're hardcoding `hints` to `None` - is this something that we need to work on in another PR?

open_feature/open_feature_client.py
```diff
@@ -152,27 +166,65 @@ def evaluate_flag_details(
...
+            evaluation_context=evaluation_context,
+            client_metadata=None,
+            provider_metadata=None,
+        )
+        merged_hooks = []
+
         try:
-            # The merge below will eventually pull in Evaluation contexts from Hooks
-            invocation_context = EvaluationContext()
+            invocation_context = before_hooks(
```
matthewelwell, line 195: I'm a bit confused by the use of the word `invocation_context` here but then `merged_context` below. Is it worth adding some comments to explain a little about what's happening with the different context objects and why?

open_feature/open_feature_api.py
matthewelwell: Ah, I think it does? This can probably be removed?

open_feature/provider/no_op_provider.py
```diff
@@ -16,7 +17,7 @@ def get_boolean_details(
         self,
         key: str,
         default_value: bool,
-        evaluation_context: typing.Any = None,
+        evaluation_context: EvaluationContext = EvaluationContext(),
```
matthewelwell, line 20: This may suffer from the same issue that I commented about before around mutable default arguments.

matthewelwell (changes requested):
We should also add some tests to the api to ensure that the hooks are run and the contexts are all merged as expected. Might be quite tricky but it would also help to serve as documentation perhaps.

open_feature/open_feature_api.py
ajhelsby: Will remove it, I must have missed it when I moved it to it's own module

open_feature/open_feature_client.py
ajhelsby: The problem with this is all the merges fail if the evaluation context is not set to something

open_feature/open_feature_client.py
ajhelsby: Yes hints is something we'll come back to in a later pr

open_feature/hooks/hook.py
ajhelsby: Yeah I originally wrote it as an optional but it seemed unessecary, I'll change the comment

open_feature/hooks/hook.py
ajhelsby: Yeah comments were taken from the java code. I'll change them.

open_feature/hooks/hook.py
ajhelsby: Yes agreed a copy and paste error

open_feature/hooks/hook.py
ajhelsby: I will add something

open_feature/hooks/hook.py
ajhelsby: Yep, added

open_feature/hooks/hook_context.py
ajhelsby: Yes agreed, only supported in python 3.7+ so if we decide to go below that we'd have to remove it

open_feature/hooks/hook_support.py
ajhelsby: If the hook does not support string flags for example we don't want to process them

open_feature/open_feature_client.py
matthewelwell: So the issue (I think) with what you had previously is that the default argument is evaluated when the module is loaded into memory, so if multiple instances of the OpenFeatureClient are instantiated without a `context` argument, they'll actually end up sharing the `EvaluationContext` object that was instantiated when the module was first loaded. Any changes made to the context in one client would also be seen in the other. 

That's my understanding at least. 

Looks like you've changed it to what I suggested anyway so I'll go ahead and resolve this conversation.

open_feature/provider/no_op_provider.py
matthewelwell: As per my comment above, we should change this to ` = None` and then instantiate the `EvaluationContext` object in the method itself.

open_feature/hooks/hook_support.py
```diff
@@ -0,0 +1,128 @@
...
+def error_hooks(
+    flag_type: FlagType,
+    hook_context: HookContext,
+    exception: Exception,
+    hooks: typing.List[Hook],
+    hints: dict,
+):
+    """
+
+    :param flag_type:
```
matthewelwell, line 22: We should either complete these docstrings or remove them

open_feature/hooks/hook.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def error(self, hook_context: HookContext, exception: Exception, hints: dict):
+        """
+        Run when evaluation encounters an error. Errors thrown will be swallowed.
+
+        :param hook_context: Information about the particular flag evaluation
+        :param exception: Ability to throw a specific exception if required
```
matthewelwell, line 43: I assumed this would be the exception that was thrown rather than the ability to throw an exception?

open_feature/open_feature_client.py
```diff
@@ -152,27 +166,65 @@ def evaluate_flag_details(
...
-                invocation_context,
+                merged_context,
                 flag_evaluation_options,
             )
+
+            after_hooks(type, hook_context, flag_evaluation, merged_hooks, None)
+
+            return flag_evaluation
         except OpenFeatureError as ofe:
+            error_hooks(flag_type, hook_context, merged_hooks, None)
```
matthewelwell, line 217: It looks like we're missing the `exception` argument from the function call here?

The definition of the function from the `hook_support` module:

```
def error_hooks(
    flag_type: FlagType,
    hook_context: HookContext,
    exception: Exception,
    hooks: typing.List[Hook],
    hints: dict,
)
```

open_feature/hooks/hook_support.py
ajhelsby: Docstrings updated

open_feature/provider/no_op_provider.py
ajhelsby: Changed to use option 3

open_feature/open_feature_client.py
ajhelsby: Comments have been added to the code and a link to the spec

open_feature/open_feature_client.py
```diff
@@ -199,6 +257,10 @@ def create_provider_evaluation(
             flag_evaluation_options,
         )
 
+        if not self.provider:
+            logging.info("No provider configured, using no-op provider.")
+            NoOpProvider()
```
matthewelwell, line 262: Should this be `self.provider = NoOpProvider()` ?

open_feature/provider/no_op_provider.py
```diff
@@ -30,7 +31,7 @@ def get_string_details(
         self,
         key: str,
         default_value: str,
-        evaluation_context: typing.Any = None,
+        evaluation_context: EvaluationContext = EvaluationContext(),
```
matthewelwell, line 34: Looks like there are still a few cases here that need reverting back to `evaluation_context: EvaluationContext = None` ?

ajhelsby: ok done

open_feature/open_feature_client.py
ajhelsby: yes, done

matthewelwell (approved):

PR #22, opened 2022-10-10
Pypi publish

ajhelsby:
## This PR
Update to the pyproject.toml configuration for publishing to pypi


pyproject.toml
```diff
@@ -4,10 +4,10 @@ requires      = ["setuptools>=61.0.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "python_open_feature_sdk"
+name = "open_feature_sdk"
 version = "0.0.1"
 description = "Standardizing Feature Flagging for Everyone"
-readme = "README.md"
+readme = "readme.md"
 authors = [{ name = "Open Feature", email = "opensource@dynatrace.com" }]
```
beeme1mr, line 11:
```suggestion
authors = [{ name = "OpenFeature", email = "openfeature-core@groups.io" }]
```

ajhelsby: Updated

.github/workflows/build-and-deploy.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        container: [ "python:3.8", "python:3.9", "python:3.10" ]
+        container: [ "python:3.10" ]
```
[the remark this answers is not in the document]
ajhelsby: Because we are deploying the package we only need to build in version of python. We only need to run the different versions of python when we are running the tests to make sure they are compatible

.github/workflows/pullrequest.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Run flake8 formatter check
+        run: flake8 .
+
+      - name: Run isort formatter check
+        run: isort .
+
+      - name: Test with pytest
+        run: coverage run --omit="*/test*" -m pytest
+
+      - name: Upload coverage to Codecov
```
matthewelwell, line 62: Does Codecov know what branch we're sending from here? Does that matter? Just making sure we don't overwrite e.g. the coverage value for the `main` branch with that of a feature branch.

requirements.txt
```diff
@@ -0,0 +1,6 @@
+#
+# This file is autogenerated by pip-compile with python 3.9
```
matthewelwell, line 2: I know it's an empty file but we should probably standardise our development environment. All the `requirements*.txt` files should be generated using a single version of python - probably 3.10 is the best option. 

Again, it doesn't really matter as long as our tests are running against all of the versions of python we want to be compatible with but just for aesthetics it would be good to be consistent here.

.github/workflows/pullrequest.yml
ajhelsby: yeah it picks up the branch from the action details, you can see this in codecov.io's dashboard

matthewelwell (approved):

requirements.txt
ajhelsby: Updated to version 3.10

Issue #7, opened 2022-08-05
Update SDK to be compliant with spec v0.2.0


beeme1mr:
Update the SDK to be compliant with the v0.2.0 spec.

## Flag Evaluation API
- [ ]  [#116](https://github.com/open-feature/spec/pull/116) - Have different methods for float and int

## Providers
- [ ]  [#119](https://github.com/open-feature/spec/pull/119) - Remove context transformers. Add provider hooks

## Evaluation Context
- [ ]  [#117](https://github.com/open-feature/spec/pull/117) - Ensure context merging behavior is correct
- [ ]  [#120](https://github.com/open-feature/spec/pull/120) - Add the ability to fetch all custom fields
- [ ]  [#121](https://github.com/open-feature/spec/pull/121) - Ensure evaluation context keys are unique

## Hooks
- [ ]  [#119](https://github.com/open-feature/spec/pull/119) - Ensure provider hook are correctly ordered

**Full Changelog**: https://github.com/open-feature/spec/compare/v0.1.0...v0.2.0

mschoenlaub:
@beeme1mr Not sure, but "Add the ability to fetch all custom fields" and "Ensure evaluation context keys are unique" are solved implicitly. One could access `ctx.attributes` and the fact that they are a `dict` should solve the uniqueness. Agreed?

beeme1mr:
Yes, I agree those requirements would be addressed implicitly with a dictionary.

PR #23, opened 2022-10-15
Add methods for integer and float flags to client

mschoenlaub:
## This PR
This PR adds methods to resolve integer and float flags to the client. They only delegate to the provider's method to retrieve `number` flags for now.

- Adds get_float_details and get_integer_details to the client

### Related Issues

Fixes a part of #7 

### Notes
This PR deliberately does not touch the `Provider` interface.

### Follow-up Tasks
Add additional methods to the `Provider` and use duck-typing to deduce whether we should convert or not.


beeme1mr:
@dlopes7 would you mind taking a look at this?

open_feature/open_feature_client.py
```diff
@@ -260,11 +318,23 @@ def create_provider_evaluation(
...
-        return get_details_callable(*args)
+        value = get_details_callable(*args)
+
+        converter = {
+            FlagType.FLOAT: float,
+            FlagType.INTEGER: int,
+        }.get(flag_type)
+
+        if converter:
+            value.value = converter(value.value)
```
tcarrio, line 338: This should probably check if the type is a mismatch. If you're expecting a float and get a string, this would be where you get the `ValueError`.

tcarrio: Additionally, if you're asking for an integer and get `3.14`, is this a failure state? You didn't get the data type you had expected.

toddbaert:
Is this PR still relevant? It seems like this might have been implemented already, unless I'm missing something! Sorry if there was duplicated effort @mschoenlaub

PR #27, opened 2022-10-18
fix/flag-error-message: Move flag evaluation details to a dataclass

ajhelsby:
## This PR
Move flag evaluation details to a dataclass and add error message to the class

- FlagEvaluationDetails has been moved to a dataclass
- error_message has been added to the FlagEvaluationDetails class

### DIFF
--- a/open_feature/flag_evaluation/flag_evaluation_details.py
+++ b/open_feature/flag_evaluation/flag_evaluation_details.py
@@ -1,18 +1,15 @@
+import typing
+from dataclasses import dataclass
+
 from open_feature.flag_evaluation.error_code import ErrorCode
 from open_feature.flag_evaluation.reason import Reason
 
 
+@dataclass
 class FlagEvaluationDetails:
-    def __init__(
-        self,
-        key: str,
-        value,
-        reason: Reason,
-        error_code: ErrorCode = None,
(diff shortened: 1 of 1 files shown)


open_feature/provider/no_op_provider.py
```diff
@@ -19,7 +19,7 @@ def get_boolean_details(
         evaluation_context: EvaluationContext = None,
     ):
         return FlagEvaluationDetails(
-            key=key,
+            flag_key=key,
```
matthewelwell, line 22: Should the argument to this function change to `flag_key` too?

ajhelsby: Not according to the spec or the other sdk's https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/FeatureProvider.java and https://github.com/open-feature/go-sdk/blob/main/pkg/openfeature/provider.go

ajhelsby: Ignore the above, agreed that the spec suggests using flag_key here https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#requirement-131 this has now been updated

matthewelwell (approved):

PR #31, opened 2022-10-29
feat: process flag evaluation options in client

tcarrio:
## This PR

- performs validation of options as dictionary before pulling hooks and hints
- introduces immutable dictionary extension of dict following naming from 3.12 release
- documented "upgrade" process for MappingProxyType

### Related Issues

Provides additional steps towards spec compliance.

### Follow-up Tasks

Hooks are completely unused still, but I'm chunking out separate PRs of work.

### How to test

Added unit tests where applicable (nothing exists yet for the client though)


tcarrio:
Signed off on commits

tcarrio:
This needs a small fix in the typing to resolve the build issues, but otherwise this is a compatible change that makes the hooks inputs actually immutable, so worthwhile to get this working again.

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+
+Python 3.8 and above are required.
+
+### Target version(s)
+
+Python 3.8 and above are supported by the SDK.
+
+### Installation and Dependencies
+
+A [`Makefile`](./Makefile) has been included in the project which should make it straightforward to start the project locally. We utilize virtual environments (see [`virtualenv`](https://docs.python.org/3/tutorial/venv.html)) in order to provide isolated development environments for the project. This reduces the risk of invalid or corrupt global packages. It also integrates nicely with Make, which will detect changes in the `requirements-dev.txt` file and update the virtual environment if any occur.
```
justinabrahms, line 15: Is using something like pipenv not common? I've been out of the python world for a few years now.. but this is what I've been using in my projects. Handles generating lock files, contains scripts, etc. 

https://pipenv.pypa.io/en/latest/index.html#

open_feature/open_feature_client.py
```diff
@@ -250,15 +251,11 @@ def evaluate_flag_details(
...
-            + flag_evaluation_options.hooks
-            + self.provider.get_provider_hooks()
+            self.hooks + evaluation_hooks + self.provider.get_provider_hooks()
         )
         # after, error, finally: Provider, Invocation, Client, API
         reversed_merged_hooks = (
-            self.provider.get_provider_hooks()
-            + flag_evaluation_options.hooks
-            + self.hooks
+            self.provider.get_provider_hooks() + evaluation_hooks + self.hooks
```
justinabrahms, line 258: I'd recommend just setting the reverse of the other one, such that it reduces the likelihood we update it in one place and not the other.

open_feature/open_feature_client.py
```diff
@@ -284,13 +281,18 @@ def evaluate_flag_details(
             )
 
             after_hooks(
-                flag_type, hook_context, flag_evaluation, reversed_merged_hooks, None
+                flag_type,
+                hook_context,
+                flag_evaluation,
+                reversed_merged_hooks,
+                hook_hints,
             )
```
justinabrahms, line 289: Seems like there are several common args among these. If you set this up with kwargs instead of args, you could do something like this, which (you make the call) could improve readability.

```python
common_hook_params = {'flag_type': ..., 'hook_context': ..., 'hook_hints': ...}
# ...
after_hooks(flag_evaluation, hook_context, **common_hook_params)
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+cd openfeature-python-sdk
+```
+
+Add your fork as an origin
+
+```bash
+git remote add fork https://github.com/YOUR_GITHUB_USERNAME/python-sdk.git
+```
+
+Makes sure your development environment is all setup by building and testing
```
StackScribe, line 55:
```suggestion
Ensure your development environment is all set up by building and testing
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+
+```bash
+make
+```
+
+To start working on a new feature or bugfix, create a new branch and start working on it.
+
+```bash
+git checkout -b feat/NAME_OF_FEATURE
```
StackScribe, line 64: Should we precede this step with steps to check out main -- I use git fetch --all --prune then git checkout upstream/main but perhaps there is a better way.

Perhaps everyone contributing is sophisticated to know this, but I've been amazed over the years at how many people don't realize they need to do this or their branch will include the changes from the branch they worked on last.

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+Open a pull request against the main js-sdk repository.
+
+### How to Receive Comments
+
+- If the PR is not ready for review, please mark it as
+  [`draft`](https://github.blog/2019-02-14-introducing-draft-pull-requests/).
+- Make sure all required CI checks are clear.
+- Submit small, focused PRs addressing a single concern/issue.
+- Make sure the PR title reflects the contribution.
+- Write a summary that helps understand the change.
```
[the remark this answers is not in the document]
StackScribe: Also,  if the PR solves an open issue, do you want that referenced in the description of the PR?

CONTRIBUTING.md
tcarrio: Either works. I haven't used `pipenv` yet but it seems like it would work well. I don't really have the time to research and change this all to use `pipenv` instead, so I created [an issue](https://github.com/open-feature/python-sdk/issues/59) to do so. If we prefer not to use Make I'll drop the code instead.

CONTRIBUTING.md
tcarrio: Since the instructions are written in the context of having just cloned the repository to start the work, so I'm not sure that adding additional steps would make sense since they are unnecessary in this context

open_feature/open_feature_client.py
```diff
@@ -359,3 +362,24 @@ def _create_provider_evaluation(
             raise TypeMismatchError()
 
         return value
+
+    def __extract_evaluation_options(
+        self, flag_evaluation_options: typing.Any
```
toddbaert, line 367: We have a type for `FlagEvaluationOptions` now.

open_feature/open_feature_client.py
tcarrio: These aren't immediately usable across all of the hooks, so while this might be able to reduce a few lines of code I don't think it's necessary to do.

open_feature/open_feature_client.py
tcarrio: Now that this type should be passed directly I'm removing the method to extract it from a dictionary.

open_feature/open_feature_client.py
tcarrio: This is done now the same way as the other SDKs.

This would technically result in different hook execution behavior if any of the hook lists has more than one entry. As it is listed in [the spec](https://docs.openfeature.dev/docs/specification/sections/hooks#requirement-442):

> Hooks **MUST** be evaluated in the following order:
>
> - before: API, Client, Invocation, Provider
> - after: Provider, Invocation, Client, API
> - error (if applicable): Provider, Invocation, Client, API
> - finally: Provider, Invocation, Client, API

The terminology in the specification does not appear to be strict enough to dictate this behavior in one such manner, such that both this Python implementation and the JavaScript one technically appease the spec but currently result in different hook behaviors in the case that any of the hook lists contain more than one hook.

I opened [an issue in the specification](https://github.com/open-feature/spec/issues/164#issue-1465298095) in order to bring clarity around the behavior of the latter 3 hooks.

toddbaert: I think regardless of how this spec issue proceeds, we can move forward with this PR. If we do make a spec change, the normal process is to review all SDKs, so it will be addressed if/when that happens.

open_feature/open_feature_client.py
toddbaert: Either I'm misunderstanding your comment, or there's a commit you haven't pushed. :sweat_smile:

toddbaert (approved):

CONTRIBUTING.md
matthewelwell: I have a chip on my shoulder regarding pipenv as it went ~2 years unmaintained and was buggy as heck. It looks like it's maintained again now but my grudge is also well maintained 😄 

If we want to move to a dedicated packaging product, I would prefer [poetry](https://github.com/python-poetry/poetry). I'll follow up on the issue you created @tcarrio .

CONTRIBUTING.md
matthewelwell: I tend to agree to with @tcarrio but thanks for the comment @StackScribe .

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+To start working on a new feature or bugfix, create a new branch and start working on it.
+
+```bash
+git checkout -b feat/NAME_OF_FEATURE
+# Make your changes
+git commit
+git push fork feat/NAME_OF_FEATURE
+```
+
+Open a pull request against the main js-sdk repository.
```
matthewelwell, line 70:
```suggestion
Open a pull request against the main python-sdk repository.
```

open_feature/open_feature_client.py
matthewelwell: Yep, I also considered this when I reviewed the original implementation. As @tcarrio says, the current implementation ensures that each hook type is implemented in an order defined by the user, within the global ordering of the different hook types. 

I agree that we should resolve this conversation and update it if / when the spec is updated.

open_feature/open_feature_client.py
```diff
@@ -359,3 +362,24 @@ def _create_provider_evaluation(
             raise TypeMismatchError()
 
         return value
+
+    def __extract_evaluation_options(
```
matthewelwell, line 366: I'm not sure if perhaps your comment on the thread beneath this one affects this, however, I'd question why this method is prefixed with a double underscore? I think you are intending to just mark this method as private (which should use a single leading underscore by convention) rather than utilise any name mangling functionality (which uses the double leading underscore).

tcarrio: Yeah this would be removed entirely, but yes it's mostly meant to be private not name mangled.

open_feature/open_feature_client.py
tcarrio: Definitely the latter- and it appears I had done that work on my desktop instead of my thinkpad so it'll be another day before I push it up.

tcarrio: Definitely the latter. I re-implemented it just now since the commit doesn't appear to have been on my thinkpad :smile:

tcarrio:
The last outstanding work for this review seems to be utilizing a different build tool besides Make. I created [a separate issue](https://github.com/open-feature/python-sdk/issues/59) to track that work.

In the meantime would we prefer to remove Make until Poetry/pipenv/etc. is implemented or leave it to be replaced when that issue is taken on?

CONTRIBUTING.md
tcarrio: I brought this up in the main thread [here](https://github.com/open-feature/python-sdk/pull/31#issuecomment-1334767068):

> The last outstanding work for this review seems to be utilizing a different build tool besides Make. I created [a separate issue](https://github.com/open-feature/python-sdk/issues/59) to track that work.
> 
> In the meantime would we prefer to remove Make until Poetry/pipenv/etc. is implemented or leave it to be replaced when that issue is taken on?

matthewelwell: I'm fine with keeping the makefile for now and dealing with the pipenv / poetry / etc. refactor later.

PR #38, opened 2022-11-02
feat: spec-0.2.0

### LINKED ISSUE #7: Update SDK to be compliant with spec v0.2.0

ajhelsby:
## This PR

Adds all the features required to reach compliance with spec 0.2.0
- Have different methods for float and int
- Remove context transformers. Add provider hooks
- Ensure context merging behavior is correct
- Add the ability to fetch all custom fields
- Ensure evaluation context keys are unique
- Ensure provider hook are correctly ordered

### Related Issues

Fixes #7

### Notes

"Add the ability to fetch all custom fields" and "Ensure evaluation context keys are unique" are solved implicitly. One could access ctx.attributes and the fact that they are a dict should solve the uniqueness.

### How to test
Unit tests 💯

### DIFF
--- a/open_feature/flag_evaluation/flag_evaluation_options.py
+++ b/open_feature/flag_evaluation/flag_evaluation_options.py
@@ -1,9 +1,14 @@
 from dataclasses import dataclass, field
+from typing import List
 
 from open_feature.hooks.hook import Hook
 
 
 @dataclass
 class FlagEvaluationOptions:
-    hooks: list[Hook] = field(default_factory=list)
+    hooks: List[Hook] = field(default_factory=list)
     hook_hints: dict = field(default_factory=dict)
+
+    def __init__(self):
+        self.hooks = []
+        self.hook_hints = {}
--- a/open_feature/flag_evaluation/flag_evaluation_options.py
+++ b/open_feature/flag_evaluation/flag_evaluation_options.py
@@ -1,14 +1,10 @@
 import typing
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from open_feature.hooks.hook import Hook
 
 
 @dataclass
 class FlagEvaluationOptions:
-    hooks: typing.List[Hook]
-    hook_hints: dict
-
-    def __init__(self):
-        self.hooks = []
-        self.hook_hints = {}
+    hooks: typing.List[Hook] = field(default_factory=list)
+    hook_hints: dict = field(default_factory=dict)
--- a/open_feature/flag_evaluation/flag_evaluation_options.py
+++ b/open_feature/flag_evaluation/flag_evaluation_options.py
@@ -1,13 +1,13 @@
-from dataclasses import dataclass, field
-from typing import List
+import typing
+from dataclasses import dataclass
 
 from open_feature.hooks.hook import Hook
 
 
 @dataclass
 class FlagEvaluationOptions:
-    hooks: List[Hook] = field(default_factory=list)
-    hook_hints: dict = field(default_factory=dict)
+    hooks: typing.List[Hook]
+    hook_hints: dict
 
     def __init__(self):
         self.hooks = []
--- a/open_feature/flag_evaluation/flag_evaluation_options.py
+++ b/open_feature/flag_evaluation/flag_evaluation_options.py
@@ -1,0 +1,9 @@
+from dataclasses import dataclass, field
+
+from open_feature.hooks.hook import Hook
+
+
+@dataclass
+class FlagEvaluationOptions:
+    hooks: list[Hook] = field(default_factory=list)
+    hook_hints: dict = field(default_factory=dict)
--- a/open_feature/open_feature_client.py
+++ b/open_feature/open_feature_client.py
@@ -205,7 +205,7 @@ class OpenFeatureClient:
         Evaluate the flag requested by the user from the clients provider.
 
         :param flag_type: the type of the flag being returned
-        :param key: the string key of the selected flag
+        :param flag_key: the string key of the selected flag
         :param default_value: backup value returned if no result found by the provider
         :param evaluation_context: Information for the purposes of flag evaluation
         :param flag_evaluation_options: Additional flag evaluation information
@@ -336,11 +336,7 @@ class OpenFeatureClient:
 
         value = get_details_callable(*args)
 
-        self._check_flag_type(flag_type, value.value)
+        if not isinstance(value.value, flag_type.value):
+            raise TypeMismatchError()
 
         return value
-
-    @staticmethod
-    def _check_flag_type(flag_type: FlagType, current_value):
-        if not isinstance(current_value, flag_type.value):
-            raise TypeMismatchError()
--- a/open_feature/open_feature_client.py
+++ b/open_feature/open_feature_client.py
@@ -1,6 +1,5 @@
 import logging
 import typing
-from numbers import Number
 
 from open_feature.evaluation_context.evaluation_context import EvaluationContext
 from open_feature.exception.exceptions import (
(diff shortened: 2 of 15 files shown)


open_feature/open_feature_client.py
```diff
@@ -272,11 +344,30 @@ def _create_provider_evaluation(
         get_details_callable = {
             FlagType.BOOLEAN: self.provider.get_boolean_details,
             FlagType.NUMBER: self.provider.get_number_details,
+            FlagType.INTEGER: self.provider.get_number_details,
```
matthewelwell, line 347: One thing I'm not clear on (from reading the spec and comparing against other sdks, e.g. [the java sdk](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/FeatureProvider.java)), is whether we should be providing all of these options backed up by a single method on the provider. 

It seems to me like we've settled in some middle ground where we're providing integer / float methods on the client but not encouraging providers to do the same. 

I would suggest that we either 'strictly' type on both sides or neither? I'm interested in other peoples' feedback here though? e.g. @justinabrahms @beeme1mr @tcarrio

open_feature/flag_evaluation/flag_evaluation_options.py
```diff
@@ -0,0 +1,10 @@
+import typing
+from dataclasses import dataclass, field
+
+from open_feature.hooks.hook import Hook
+
+
+@dataclass
+class FlagEvaluationOptions:
```
matthewelwell, line 8: I can't see this in the [list of items for spec 0.2.0](https://github.com/open-feature/python-sdk/issues/7)? Is this something from a later update to the spec - or it was just missing from the GH issue?

open_feature/open_feature_client.py
```diff
@@ -190,7 +258,11 @@ def evaluate_flag_details(
             client_metadata=None,
             provider_metadata=None,
         )
-        merged_hooks = self.hooks
+        merged_hooks = (
```
matthewelwell, line 261: It seems to me that the hooks should be in different orders depending on the stage in the evaluation lifecycle, whereas I think we're just defining a single order and passing these to each `x_hooks` function call?

https://github.com/open-feature/spec/blob/main/specification/sections/04-hooks.md#requirement-442

Specifically, it looks like before_hooks are evaluated in the opposite order to the rest of the hook evaluations.

open_feature/flag_evaluation/flag_evaluation_options.py
ajhelsby: Flag evaluation options are needed for the hook providers being handled in the correct order

open_feature/open_feature_client.py
ajhelsby: JS: https://github.com/open-feature/js-sdk/blob/main/src/no-op-provider.ts Using generic number

Java: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/FeatureProvider.java Specific number types

toddbaert: (Non-python engineer, so take that into account) I think I agree with @matthewelwell that it should be consistent. As far as if integer/number should be individually supporting, consider:

- support for fractional numbers could be important, especially <1. I could see feature flags that are proportions quite easily (e.g. 0.5)
- precision loss with high numbers (probably less important in feature flagging use-cases)

open_feature/open_feature_client.py
ajhelsby: I've put in a new order for the other hooks

open_feature/open_feature_client.py
matthewelwell: I'm inclined to lean towards the Java approach here. It feels like the generic `number` approach suits JS better with the closer ties to JSON but in python terms 'number' doesn't really mean an awful lot. I'd suggest that we standardise to having 2 methods, one for int and one for float, both on the client and on the provider. 

That said, I'm open to being disagreed with here 😄

open_feature/flag_evaluation/flag_evaluation_options.py
matthewelwell: Understood.

open_feature/open_feature_client.py
ajhelsby: Agreed I would always prefer stronger typing if possible. Python seems to be moving that way, better to future proof it.

open_feature/open_feature_client.py
```diff
@@ -271,12 +325,30 @@ def _create_provider_evaluation(
...
             FlagType.STRING: self.provider.get_string_details,
         }.get(flag_type)
 
         if not get_details_callable:
             raise GeneralError(error_message="Unknown flag type")
 
-        return get_details_callable(*args)
+        value = get_details_callable(*args)
+
+        if flag_type in NUMERIC_TYPES:
```
matthewelwell, line 339: I don't think we need to do this anymore, right? We can just rely on the provider providing the correct type from their methods?

open_feature/open_feature_client.py
```diff
@@ -271,12 +325,30 @@ def _create_provider_evaluation(
...
+            value.value = self._convert_numeric_types(flag_type, value.value)
+
+        return value
+
+    @staticmethod
+    def _convert_numeric_types(flag_type: FlagType, current_value):
```
matthewelwell, line 345: As above, I think this method can be entirely removed.

open_feature/open_feature_client.py
ajhelsby: Added an `isinstance` check instead

open_feature/open_feature_client.py
beeme1mr: I'm also weak on the Python front but I'm all for stronger type safety. As long as a Python dev wouldn't be confused or put off by this, I would say go for it.

matthewelwell (approved):

toddbaert (commented):
Nice job! I have virtually no python experience, but I've left some comments for your consideration.

:pray:

open_feature/provider/no_op_provider.py
```diff
@@ -40,9 +45,22 @@ def get_string_details(
             variant=PASSED_IN_DEFAULT,
         )
 
-    def get_number_details(
+    def get_integer_details(
         self,
-        flag_key: str,
+        flag_key: int,
+        default_value: Number,
```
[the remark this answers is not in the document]
ajhelsby: Oops yes wrong parameter, fixed now

open_feature/provider/no_op_provider.py
```diff
@@ -40,9 +45,22 @@ def get_string_details(
...
+        return FlagEvaluationDetails(
+            flag_key=flag_key,
+            value=default_value,
+            reason=Reason.DEFAULT,
+            variant=PASSED_IN_DEFAULT,
+        )
+
+    def get_float_details(
+        self,
+        flag_key: float,
```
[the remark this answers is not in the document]
ajhelsby: Same as above

open_feature/open_feature_client.py
```diff
@@ -99,30 +105,58 @@ def get_string_details(
             flag_evaluation_options,
         )
 
-    def get_number_value(
+    def get_integer_value(
+        self,
+        flag_key: str,
+        default_value: int,
+        evaluation_context: EvaluationContext = None,
+        flag_evaluation_options: typing.Any = None,
```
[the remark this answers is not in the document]
ajhelsby: Updated to use the FlagEvaluationOptions dataclass

matthewelwell:
@toddbaert thanks for the feedback! @ajhelsby has made the changes you mentioned (that I missed 😬), I'm good to merge this now if you are happy with it?

toddbaert (approved):
lgtm! Thank you!

PR #43, opened 2022-11-03
eval context fixes and new error types

rgrassian-split:
Signed-off-by: Robert Grassian <robert.grassian@split.io>

## This PR
- Changes the evaluation context merge to create and return a new object rather than modify itself. This fixes a bug where the openfeature api level context was being overwritten by the client's context
- Creates `TargetingKeyMissingError` and `InvalidContextError`, along with their error codes
- Adds an evaluation context setter on the client


open_feature/open_feature_client.py
```diff
@@ -280,3 +280,6 @@ def _create_provider_evaluation(
             raise GeneralError(error_message="Unknown flag type")
 
         return get_details_callable(*args)
+
+    def set_evaluation_context(self, evaluation_context: EvaluationContext):
```
matthewelwell, line 284: Is there a specific reason we need this setter method? Feels like a bit of a python anti-pattern since one could just set it directly on the object? We also already have the evaluation context in the constructor so I'm unclear what this method gives us? 

If there's a good reason to keep it, however, I'm keen to structure the methods slightly, primarily just keeping public methods above any private methods.

open_feature/exception/exceptions.py
```diff
@@ -85,3 +85,37 @@ def __init__(self, error_message: str = None):
...
+    but one was not provided in the evaluation context.
+    """
+
+    def __init__(self, error_message: str = None):
+        """
+        Constructor for the TargetingKeyMissingError. The error code for this type of
+        exception is ErrorCode.TARGETING_KEY_MISSING.
+        @param error_message: a string message representing why the error has been
+        raised
+        @return: the generic TargetingKeyMissingError exception
```
matthewelwell, line 102: I've just noticed that this is a general issue with all the exceptions in this module... but technically this method doesn't actually return anything, would you mind removing the returns on the new exceptions you've added and I'll get the rest cleaned up.

open_feature/exception/exceptions.py
```diff
@@ -85,3 +85,37 @@ def __init__(self, error_message: str = None):
...
+        """
+        super().__init__(error_message, ErrorCode.TARGETING_KEY_MISSING)
+
+
+class InvalidContextError(OpenFeatureError):
+    """
+    This exception should be raised when The evaluation context does not meet provider
```
matthewelwell, line 109: Being very picky now but there's a minor typo here

```suggestion
    This exception should be raised when the evaluation context does not meet provider
```

matthewelwell (commented):
Thanks @rgrassian-split, totally agree with the new approach for the context merging, thanks for submitting this. I've added some comments in a few places.

open_feature/evaluation_context/evaluation_context.py
```diff
@@ -7,7 +7,7 @@ def merge(self, ctx2: "EvaluationContext") -> "EvaluationContext":
         if not (self and ctx2):
             return self or ctx2
 
-        self.attributes = {**self.attributes, **ctx2.attributes}
-        self.targeting_key = ctx2.targeting_key or self.targeting_key
+        attributes = {**self.attributes, **ctx2.attributes}
+        targeting_key = ctx2.targeting_key or self.targeting_key
 
-        return self
+        return EvaluationContext(targeting_key=targeting_key, attributes=attributes)
```
ajhelsby, line 13: 👍🏻

open_feature/open_feature_client.py
rgrassian-split: The typical way a client is created is not though the constructor but through the `openFeatureAPI.get_client()` method. I agree its not necessary though and we can just set the value directly, unless we want it to be more similar to other sdks. I'm happy to remove it though.

open_feature/exception/exceptions.py
rgrassian-split: Yep sounds good

open_feature/exception/exceptions.py
rgrassian-split: Fixed

matthewelwell (approved):

open_feature/open_feature_client.py
matthewelwell: Ah yes, that's a good point. I see that you removed it already - lets leave it out and then if we find that it would be useful to have in the future we can add it in at a later stage. Thanks!

beeme1mr:
@matthewelwell @ajhelsby any concerns about merging this?

PR #45, opened 2022-11-11
feat: Add release please to handle releases

ajhelsby:
## This PR

Add the release please action to handle release versioning


.release-please-manifest.json
```diff
@@ -0,0 +1 @@
+{".":"1.0.0"}
```
beeme1mr, line 1: Please change this to the current version.

.github/workflows/release.yml
```diff
@@ -23,34 +26,49 @@ jobs:
...
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          command: manifest
+          token: ${{secrets.GITHUB_TOKEN}}
+          default-branch: main
+          release-type: python
+
       - uses: actions/checkout@v3
+        if: ${{ steps.release.outputs.releases_created }}
```
beeme1mr, line 38: This approach works but it may be a bit easier to manage if you break release please and the actual release process into multiple jobs. That way, you only need to check for a release once. Here's an example:

https://github.com/open-feature/js-sdk/blob/main/.github/workflows/release-please.yml#L25

beeme1mr (commented):
You may want to consider adding a `release-please-config.json` to the root of the project. You can reuse most of the settings found [here](https://github.com/open-feature/js-sdk/blob/main/release-please-config.json). You'll just need to change the release type and bootstrap sha.

Finally, you can use release please to bump versions in the readme. You just need to add them to the config like [this ](https://github.com/open-feature/open-feature-operator/blob/main/release-please-config.json#L9) and add comments to the files that need to have a version bumped on release. You can see an example [here](https://github.com/open-feature/open-feature-operator/blob/main/README.md?plain=1#L34-L41).

Please let me know if you have any questions.

.release-please-manifest.json
ajhelsby: 👍🏻

.github/workflows/release.yml
ajhelsby: Moved the release please work out into it's own job

.github/workflows/release.yml
```diff
@@ -14,8 +14,28 @@ permissions:  # added using https://github.com/step-security/secure-workflows
...
 
 jobs:
-  build:
+  release-please:
+    permissions:
+      contents: write  # for google-github-actions/release-please-action to create release commit
+      pull-requests: write  # for google-github-actions/release-please-action to create release PR
     runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.container }}
```
beeme1mr, line 23: I don't think a container matrix is necessary for this job.

beeme1mr (commented):
I think this should work. The last part missing is the release please comments in the readme. You can see an example [here](https://github.com/open-feature/open-feature-operator/blob/main/README.md?plain=1#L34-L41). This part is required or it won't automatically bump the version in the readme as part of the release process.

beeme1mr:
It's also worth mentioning that Release Please works off semantic commits. Most of the repos use a [PR title validator](https://github.com/open-feature/js-sdk/blob/main/.github/workflows/lint-pr.yml) and perform squash merges. I'm happy to help set that up if you're interested.

ajhelsby:
> It's also worth mentioning that Release Please works off semantic commits. Most of the repos use a [PR title validator](https://github.com/open-feature/js-sdk/blob/main/.github/workflows/lint-pr.yml) and perform squash merges. I'm happy to help set that up if you're interested.

I believe this has already been set up here https://github.com/open-feature/python-sdk/blob/main/.github/workflows/lint-pr.yml

.github/workflows/release.yml
ajhelsby: Agreed, I've removed the container

beeme1mr (approved):

toddbaert (approved):
LGTM as well, but these things usually take some troubleshooting :crossed_fingers:

PR #66, opened 2022-12-27
feat: defaults to NoOpProvider

tcarrio:
## This PR

- initial provider set to a NoOpProvider
- ensures a client will always be returned even if a provider has not been explicitly set

### Related Issues

Fixes #1234523

### Notes

This fixes an issue with spec compliance

### Follow-up Tasks

~~- Tests~~

### How to test

Given you have not set a provider
When you retrieve an OpenFeature client
Then an error will not be raised

### DIFF
--- a/tests/test_open_feature_api.py
+++ b/tests/test_open_feature_api.py
@@ -6,7 +6,7 @@ from open_feature.open_feature_api import get_client, get_provider, set_provider
 from open_feature.provider.no_op_provider import NoOpProvider
 
 
-def test_should_not_raise_exception_with_nop_client():
+def test_should_not_raise_exception_with_noop_client():
     # Given
     # No provider has been set
     # When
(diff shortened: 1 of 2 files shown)


tcarrio:
Updated tests :+1:

tcarrio:
The `-0.03%` test coverage is because I removed code but had the same coverage of that file :smile:

tests/test_open_feature_api.py
```diff
@@ -6,18 +6,16 @@
 from open_feature.provider.no_op_provider import NoOpProvider
 
 
-def test_should_raise_exception_with_nop_client():
+def test_should_not_raise_exception_with_nop_client():
```
beeme1mr, line 9:
```suggestion
def test_should_not_raise_exception_with_noop_client():
```

beeme1mr (approved):
Added one comment but overall looks good.

PR #103, opened 2023-04-20
chore: split eval and resolution details objects, adjust providers accordingly

### LINKED ISSUE #97: separate `EvaluationDetails` from `ResolutionDetails`, https://openfeature.dev/specification/types/#evaluation-details

mowies:
### This PR
- adds a new `FlagResolutionDetails` class according to the spec
- makes use of the new class in the abstract provider and the no_op provider instead of using `EvaluationDetails`

Fixes #97

### DIFF
--- a/open_feature/flag_evaluation/resolution_details.py
+++ b/open_feature/flag_evaluation/resolution_details.py
@@ -1,0 +1,17 @@
+import typing
+from dataclasses import dataclass
+
+from open_feature.exception.error_code import ErrorCode
+from open_feature.flag_evaluation.reason import Reason
+
+T = typing.TypeVar("T", covariant=True)
+
+
+@dataclass
+class FlagResolutionDetails(typing.Generic[T]):
+    value: T
+    error_code: typing.Optional[ErrorCode] = None
+    error_message: typing.Optional[str] = None
+    reason: typing.Optional[Reason] = None
+    variant: typing.Optional[str] = None
+    flag_metadata: typing.Optional[str] = None
--- a/open_feature/provider/no_op_provider.py
+++ b/open_feature/provider/no_op_provider.py
@@ -1,7 +1,6 @@
 import typing
 
 from open_feature.evaluation_context.evaluation_context import EvaluationContext
-from open_feature.flag_evaluation.flag_evaluation_details import FlagEvaluationDetails
 from open_feature.flag_evaluation.reason import Reason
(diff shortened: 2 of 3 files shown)


open_feature/provider/no_op_provider.py
```diff
@@ -1,8 +1,8 @@
 import typing
 
 from open_feature.evaluation_context.evaluation_context import EvaluationContext
-from open_feature.flag_evaluation.flag_evaluation_details import FlagEvaluationDetails
 from open_feature.flag_evaluation.reason import Reason
+from open_feature.flag_evaluation.resolution_details import FlagResolutionDetails
```
thisthat, line 5: to armonize the import order, we could introduce [black](https://pypi.org/project/black/)

open_feature/flag_evaluation/resolution_details.py
```diff
@@ -0,0 +1,17 @@
...
+T = typing.TypeVar("T", covariant=True)
+
+
+@dataclass
+class FlagResolutionDetails(typing.Generic[T]):
+    value: T
+    error_code: typing.Optional[ErrorCode] = None
+    error_message: typing.Optional[str] = None
+    reason: typing.Optional[Reason] = None
+    variant: typing.Optional[str] = None
```
thisthat, line 16: I would expect variant to always be defined
```suggestion
    variant: str
```

open_feature/provider/no_op_provider.py
mowies: I ran the file through black and nothing changed 🤷🏼

open_feature/flag_evaluation/resolution_details.py
mowies: According to the spec, this is optional: https://openfeature.dev/specification/types/#resolution-details

open_feature/provider/no_op_provider.py
thisthat: Sorry, I didn't explain properly. What I was suggesting is to have a check that is run on a PR level to make sure the code is formatted properly. Something like the following: https://github.com/open-telemetry/build-tools/blob/main/.github/workflows/semconvgen.yml#L28

thisthat (approved):
LGTM

open_feature/provider/no_op_provider.py
mowies: Ah yes that makes sense, but probably in a separate issue :)

PR #134, opened 2023-07-02
fix: object client methods can receive dict or list as default_value

federicobond:
## This PR

Updates client object methods to accept dict or list as default value.

### DIFF
--- a/open_feature/open_feature_client.py
+++ b/open_feature/open_feature_client.py
@@ -182,7 +182,7 @@ class OpenFeatureClient:
     def get_object_value(
         self,
         flag_key: str,
-        default_value: dict,
+        default_value: typing.Union[dict, list],
         evaluation_context: typing.Optional[EvaluationContext] = None,
         flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None,
     ) -> dict:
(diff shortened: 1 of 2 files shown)


hlipsig (commented):
LGTM I don't see anything wrong with it from a code perspective.  This change is straighforward and easy to understand. I'm double checking there's not any obscure reason in our codebase that this shouldn't be done.

Kavindu-Dodan:
@federicobond  can we add few test cases to cover this scenario ? This will help to validate this change as well as not break this with future changes

federicobond:
@Kavindu-Dodan sure! Not sure what kind of tests are you referring to, but I did update the one I found that was missing the list case.

federicobond:
To get proper type testing we would have to run mypy or some other typechecker in some example code, but I think this exceeds the scope of this pull request.

tcarrio (approved):

federicobond:
Replaced get_object_value for get_object_details in test case. Everything should be green now.

toddbaert (approved):

Kavindu-Dodan:
> @Kavindu-Dodan sure! Not sure what kind of tests are you referring to, but I did update the one I found that was missing the list case.

Sorry for not being specific. I was referring to the handling list type and seems you already added that.

Kavindu-Dodan (approved):

PR #136, opened 2023-07-06
fix: return correct type object in OpenFeatureClient._create_provider_evaluation

federicobond:
The `_create_provider_evaluation` method is declared as returning a `FlagEvaluationDetails` but actually returned the `FlagResolutionDetails` directly from the provider.

With this change, a `FlagEvaluationDetails` object is created from the resolution details aligning the implementation with the spec and the declared types.

Similar code can found in the Java SDK:

https://github.com/open-feature/java-sdk/blob/9ddc9f1cb2c85c2d096a493342e429120ff36e92/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L134


toddbaert (approved):
Yep, looks like this was missed when we addressed [this](https://github.com/open-feature/python-sdk/issues/97).

PR #138, opened 2023-07-07
refactor: make every get_<type>_value function delegate to get_<type>_details

federicobond:
## This PR

Makes sure that each `get_<type>_value` method in OpenFeatureClient delegates to the corresponding `get_<type>_details` method. Some of these methods already did so but others called `evaluate_flag_details` directly.

To adhere to a principle of least surprise, remaining methods were adapted to follow a common calling convention.


beeme1mr:
I'll rerun the GitHub action.

beeme1mr (approved):

PR #142, opened 2023-07-11
refactor: make Reason enum values instances of str

federicobond:
## This PR

Is a refinement on #126. In particular, it makes Reason values also instances of str, and thus directly comparable without having to access the `.value` property, and without having to define custom getters and setters in dataclasses.

The motivation for this was to fix a `no-redef` issue reported by mypy in the FlagEvaluationDetails definition:

```
open_feature/flag_evaluation/flag_evaluation_details.py:19: error: Name "reason" already defined on line 15  [no-redef]
```

`StrEnum` is available since Python 3.11 so it requires a minimal backport for earlier Python versions.

### DIFF
--- a/open_feature/_backports/strenum.py
+++ b/open_feature/_backports/strenum.py
@@ -1,0 +1,11 @@
+try:
+    from enum import StrEnum
+except ImportError:
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        """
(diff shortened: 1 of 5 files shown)


beeme1mr (approved):

hlipsig (approved):
I don't have any concerns here. This is a good change.

PR #157, opened 2023-07-18
feat: implement InMemoryProvider

federicobond:
## This PR

Implements an in-memory provider that roughly follows @Kavindu-Dodan proposal in https://github.com/open-feature/ofep/pull/71

### Notes

A slightly different set of tradeoffs were made regarding the design of the InMemoryFlag object, which should be reviewed carefully. In particular, the State field in the Go implementation was replaced by an explicit Reason which seemed more ergonomic.

### DIFF
--- a/open_feature/provider/in_memory_provider.py
+++ b/open_feature/provider/in_memory_provider.py
@@ -20,7 +20,7 @@ class InMemoryMetadata(Metadata):
 T = typing.TypeVar("T", covariant=True)
 
 
-@dataclass
+@dataclass(frozen=True)
 class InMemoryFlag(typing.Generic[T]):
     flag_key: str
     default_variant: str
@@ -58,7 +58,7 @@ class InMemoryProvider(AbstractProvider):
     _flags: FlagStorage
 
     def __init__(self, flags: FlagStorage):
-        self._flags = flags
+        self._flags = flags.copy()
 
     def get_metadata(self) -> Metadata:
         return InMemoryMetadata()
--- a/open_feature/provider/in_memory_provider.py
+++ b/open_feature/provider/in_memory_provider.py
@@ -1,0 +1,122 @@
+from dataclasses import dataclass
+import typing
+
+from open_feature.evaluation_context.evaluation_context import EvaluationContext
+from open_feature.exception.error_code import ErrorCode
+from open_feature.flag_evaluation.reason import Reason
+from open_feature.flag_evaluation.resolution_details import FlagResolutionDetails
+from open_feature.hooks.hook import Hook
+from open_feature.provider.metadata import Metadata
+from open_feature.provider.provider import AbstractProvider
+
+PASSED_IN_DEFAULT = "Passed in default"
+
+
+@dataclass
+class InMemoryMetadata(Metadata):
+    name: str = "In-Memory Provider"
+
+
+T = typing.TypeVar("T", covariant=True)
+
+
+@dataclass
+class InMemoryFlag(typing.Generic[T]):
+    flag_key: str
+    default_variant: str
+    variants: typing.Dict[str, T]
+    reason: typing.Optional[Reason] = Reason.STATIC
+    error_code: typing.Optional[ErrorCode] = None
+    error_message: typing.Optional[str] = None
+    context_evaluator: typing.Optional[
+        typing.Callable[["InMemoryFlag", EvaluationContext], FlagResolutionDetails[T]]
+    ] = None
+
+    def resolve(
+        self, evaluation_context: typing.Optional[EvaluationContext]
+    ) -> FlagResolutionDetails[T]:
+        if self.context_evaluator:
+            return self.context_evaluator(
+                self, evaluation_context or EvaluationContext()
+            )
+
+        return FlagResolutionDetails(
+            value=self.variants[self.default_variant],
+            reason=self.reason,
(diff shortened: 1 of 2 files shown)


open_feature/provider/in_memory_provider.py
```diff
@@ -0,0 +1,120 @@
...
+
+FlagStorage = typing.Dict[str, InMemoryFlag]
+
+V = typing.TypeVar("V")
+
+
+class InMemoryProvider(AbstractProvider):
+    _flags: FlagStorage
+
+    def __init__(self, flags: FlagStorage):
```
federicobond, line 58: My worry with this API is that it's very tempting for users to retain a copy of the mutable data structure behind our backs, which could cause difficult to debug bugs.

Might be a bigger problems in languages without a GIL.

toddbaert: could we clone the object so that such mutation isn't possible?

open_feature/provider/in_memory_provider.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@dataclass
+class InMemoryFlag(typing.Generic[T]):
+    flag_key: str
+    default_variant: str
+    variants: typing.Dict[str, T]
+    reason: typing.Optional[Reason] = Reason.STATIC
+    error_code: typing.Optional[ErrorCode] = None
+    error_message: typing.Optional[str] = None
```
Kavindu-Dodan, line 30: I don't think we should embed error code & message into the flag structure. These values should be resolved from evaluation logic 

For example, consider this go example [1]. This is the in-memory search operation and if flag is not found based on flag key, this returns a flag not found [2].


[1] - https://github.com/open-feature/go-sdk/pull/192/files#diff-ebcfeaa6fba781be983c41afae4094c9324c532721a8bbb7b6a03cebaf37d027R132-R143
[2] - https://github.com/open-feature/spec/blob/main/specification/types.md#error-code

tests/provider/test_in_memory_provider.py
```diff
@@ -0,0 +1,154 @@
...
+    provider = InMemoryProvider(
+        {"Key": InMemoryFlag("Key", "obj", {"obj": return_value, "empty": {}})}
+    )
+    # When
+    flag = provider.resolve_object_details(flag_key="Key", default_value={})
+    # Then
+    assert flag is not None
+    assert flag.value == return_value
+    assert isinstance(flag.value, dict)
+    assert flag.variant == "obj"
```
Kavindu-Dodan, line 154: Good to see tests 🙌 


We need tests for error scenarios. For example, evaluating a flag that's missing (resolved to default with error details) [1], evaluating wrong typed flag (mixing string flag with integer .. etc). Those tests and error handling will make this complete :) 

[1] - https://github.com/open-feature/go-sdk/pull/192/files#diff-60841eaba9b2e1a73e7f57ceebe625fb14c660897d59f7289bd3ce39d05b2342R169-R188

federicobond: 1.  Handling default evaluation for missing flags is already tested in `test_should_handle_unknown_flags_correctly`
 2.  The typechecking logic in the Python SDK is implemented in the OpenFeatureClient. Would adding additional typechecking logic in the provider duplicate efforts?

open_feature/provider/in_memory_provider.py
Kavindu-Dodan: @federicobond my initial thought of in-memory provider was mainly on tests. However, some discussions and community opinions on OFEP [1] [2] suggests that this can live outside of tests. 

With that background, I think we should have some sort of basic error handling built-in to this provider 

[1] - https://github.com/open-feature/ofep/pull/71#discussion_r1263230706
[2] - https://github.com/open-feature/ofep/pull/71#issuecomment-1638384086

tests/provider/test_in_memory_provider.py
Kavindu-Dodan: 1. Thanks for the reference. Connected to comment https://github.com/open-feature/python-sdk/pull/157/files#r1266982206,  we can improve the error handling logic and make this provider behave similar to a real one
2. Interesting, didn't know that this was handled at client level in Python SDK. For example in GO [1] & Java [2] this is not validated at Client. May be it's a good idea to add a simple test and see hoe SDK behaves. Also, this will anyway get covered later by E2E tests 

[1] - https://github.com/open-feature/go-sdk/blob/main/pkg/openfeature/client.go#L709-L732
[2] - https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L160C1-L179C1

federicobond: My mistake: there are actually two typechecking errors involved here.

 * On request: when a client attempts to resolve a flag with a method that does not match the flag type
 * On response: when a provider resolves to a flag value that does not match the expected flag type

The response case is checked in OpenFeatureClient which raises a TypeMismatchError.

But the request case should probably be checked by the provider which can either:

 1. Return a FlagResolutionDetails object with the appropriate TYPE_MISMATCH error code.
 2. Raise TypeMismatchError like the OpenFeatureClient

The [specification](https://openfeature.dev/specification/sections/providers#requirement-227) is not clear about which path we should take.

beeme1mr (approved):

PR #164, opened 2023-08-31
refactor!: move api evaluation context methods to api module

federicobond:
Changes the import path for evaluation context methods from:
```python
from open_feature.open_feature_evaluation_context import api_evaluation_context, set_api_evaluation_context
```
To:
```python
from open_feature.open_feature_api import get_evaluation_context, set_evaluation_context
```

This brings it closer to what users would expect after reading the spec. We should do something similar with the api hooks methods.

I will submit a series of pull requests later to continue adapting the module structure and make it more Pythonic.


beeme1mr (approved):

beeme1mr:
Looks good to me, thanks. I've updated the PR title so that the change log marks this as a breaking change.

PR #169, opened 2023-08-31
refactor!: move api hooks methods to api module

federicobond:
This change does something similar to https://github.com/open-feature/python-sdk/pull/164 but for the hooks API. Imports should be updated from:

```python
from open_feature.hooks import api_hooks, add_api_hooks
```

To:
```python
from open_feature.open_feature_api import get_hooks, add_hooks
```

This ensures all API-level methods are accessible from the same module.


open_feature/open_feature_api.py
```diff
@@ -2,6 +2,7 @@
 
 from open_feature.evaluation_context.evaluation_context import EvaluationContext
 from open_feature.exception.exceptions import GeneralError
+from open_feature.hooks.hook import Hook
```
beeme1mr, line 5: This makes sense to me but the other imports tend to go `open_feature`, singular (e.g. exception), plural (e.g. exceptions). Personally, I prefer how you've done it here but it would be nice if they're consistent.

beeme1mr (approved):

open_feature/open_feature_api.py
federicobond: Sure! I didn't change the module name though, it was `hooks` before and it remains `hooks` now. This import was added because `open_feature_api` now needs to import Hook.

Issue #101, opened 2023-04-19
Review public APIs (make sure they are "Pythonic")


toddbaert:
We know some of the internals of the SDK are a bit "Java-y"; these can be refactored later. However, before a 1.0 we need to make sure the public APIs are idiomatic to ensure we don't regret or design choices post-1.0.

Definition of done:

- [ ] Review SDK, issues created for public API changes if necessary.

federicobond:
The biggest issue is see here is that the module structure is too deeply nested. I have no hard data on this, but in my experience one should be able to import 80-90% of the code that is needed for most use cases from `open_feature.<xyz>` and only very specialized functionality from `open_feature.<xyz>.<abc>`

Some examples:

| Original import | Proposed import |
| --- | ----------- |
| from open_feature.hooks.hook import Hook | from open_feature.hooks import Hook |
| from open_feature.provider.no_op_provider import NoOpProvider | from open_feature.provider import NoOpProvider |

If needed, additional internal structure can be handled via underscore-prefixed modules to signal that you should not import these directly.

The other Java-ism I am seeing is the `open_feature.open_feature_<x>` module naming, closely following the main class name in each case. A more Pythonic approach would be to rename them to `open_feature.client` and `open_feature.api`.

toddbaert:
@federicobond the above sounds right to me.

Because this is an issue of style, I'd love to hear some concensus from others. cc @hlipsig @matthewelwell @jamescarr @tcarrio  @open-feature/sdk-python-approvers @open-feature/sdk-python-maintainers

hlipsig:
Agreed.

beeme1mr:
Sounds like a nice change to me 👍.

matthewelwell:
Yep, agree.

gruebel:
@federicobond should we close this?

federicobond:
Yes, I find no more directly actionable issues here.

PR #172, opened 2023-09-01
refactor!: simplify namespaces to make public API more pythonic

federicobond:
For reviewing, it's best to follow the changes commit by commit.

This reduces the number of public namespaces in the sdk, following what we agreed on https://github.com/open-feature/python-sdk/issues/101

I consolidated some modules into a single file when it made sense. The alternative would have been to prefix the internal namespace files with underscores and re-export the relevant symbols from `__init__.py` but it added a lot of boilerplate with little benefit. If the modules grow in size to make this necessary, it can be done later without breaking compatibility with existing users.


beeme1mr (approved):
This looks great, thank you! Please make sure to update the examples on the readme.

toddbaert:
I'll review this first thing next week. Thanks @federicobond

tests/provider/test_in_memory_provider.py
```diff
@@ -1,8 +1,7 @@
 from numbers import Number
 
-from open_feature.exception.error_code import ErrorCode
-from open_feature.flag_evaluation.reason import Reason
-from open_feature.flag_evaluation.resolution_details import FlagResolutionDetails
+from open_feature.exception import ErrorCode
+from open_feature.flag_evaluation import FlagResolutionDetails, Reason
```
toddbaert, line 4: Ya these all feel A LOT better.

toddbaert:
> If the modules grow in size to make this necessary, it can be done later without breaking compatibility with existing users.

:+1:

toddbaert (approved):
LGTM!

toddbaert:
Obviously this is breaking, but I have no issue merging it. I'll do so in the next few days unless I hear objections from the other (more Python-literate) reviewers.

PR #215, opened 2023-10-18
refactor: use if clauses in list comprehensions to make code more pythonic

federicobond:
## This PR

Rewrites the _execute_hooks* functions to be more pythonic. This improves readability and makes the code faster since it avoids the creation of intermediate structures.

I also removed the null guards since the functions are typed as receiving lists always.


beeme1mr (commented):
Honestly, I'm not entirely sure what's happening in this change 😅. Perhaps someone with some more Python experience can chime in.

@justinabrahms @tcarrio @jamescarr

beeme1mr (approved):
The tests pass and this is just an internal refactor.

PR #216, opened 2023-10-18
fix: Hook methods should have default non-abstract implementations

federicobond:
## This PR

Fixes Hook methods incorrectly marked as abstract. The Hook class provides default implementations of all hook lifecycle methods. Hook subclasses only need to override the methods they are interested in.


openfeature/hook/__init__.py
```diff
@@ -27,8 +26,9 @@ class HookContext:
 
 
 class Hook:
-    @abstractmethod
-    def before(self, hook_context: HookContext, hints: dict) -> EvaluationContext:
+    def before(
+        self, hook_context: HookContext, hints: dict
+    ) -> typing.Optional[EvaluationContext]:
```
beeme1mr, line 31: Could you please add a test that adds evaluation context in the before hook?

beeme1mr (commented):
Why did the return value

federicobond:
@beeme1mr the `return None` change is just to comply explicitly with the declared return type of the method. It does not introduce any behavior changes and is functionally identical to the previous code.

openfeature/hook/__init__.py
federicobond: Just added a unit test for full context evaluation merging which was missing.

PR #220, opened 2023-10-19
chore: optimize GitHub CI workflow

federicobond:
## This PR

Optimizes our GH workflows by running linters like black/flake8 and CodeQL in a separate job with a single Python version, drastically reducing the feedback time for builds. Tests still run on a version matrix.

It also removes the virtualenv configuration which is not necessary in CI since the machines are ephemeral anyway, and uses the official Python setup actions instead of Docker to further reduce the number of layers involved in the execution of the build.

### TODO:

 - [ ] once approved, repeat the same procedure for the python-sdk-contrib repo.


beeme1mr (approved):
The changes look good but do you need two separate files? I believe you should be able to update one of the files to support either trigger. It also may not be necessary to run at all on a merge to main since it was tested during the PR process.

federicobond:
Indeed, I just didn't want to change too much before validating the approach. Will update the PR.

.github/workflows/build.yml
```diff
@@ -3,42 +3,62 @@
 # separate terms of service, privacy policy, and support
 # documentation.
 
-name: PR
+name: "Build"
```
beeme1mr, line 6:
```suggestion
name: "Build, lint, and test"
```

beeme1mr (approved):

federicobond:
Not sure where those missing build status come from 🤔

toddbaert:
> Not sure where those missing build status come from 🤔

I can't quite understand it either. I wonder if it's just an artifact of the changes and once it's merged it will work subsequently?

toddbaert (approved):

federicobond:
@toddbaert the problem is that it does not let me merge without every status check passing. I will open a separate PR to see if that solves it.

federicobond:
@toddbaert I think it MIGHT have to do with some of these checks being marked as required and GH not preserving the mapping from the previous configuration. Do you have permission to disable them temporarily and then enable the new ones?

beeme1mr:
@federicobond we had branch protection enabled that look for the old job names.

PR #228, opened 2023-11-01
fix!: raise error if the flag wasn't found using the in-memory provider

beeme1mr:
## This PR

- Updated the in-memory provider to throw if the flag isn't found
- Removed duplicate flag keys from the in-memory provider flag config

### Notes

I marked this as a breaking fix because it changes the method signature of the in-memory provider. Since we're sub-1.0, this will be treated as a minor version change, but highlight the change more predominately in the change log.

I originally intended to only remove the duplicate keys in this PR. If you would prefer that I split these into two separate PRs, please let me know.

### DIFF
--- a/openfeature/provider/in_memory_provider.py
+++ b/openfeature/provider/in_memory_provider.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass, field
 
 from openfeature._backports.strenum import StrEnum
 from openfeature.evaluation_context import EvaluationContext
-from openfeature.exception import ErrorCode
+from openfeature.exception import FlagNotFoundError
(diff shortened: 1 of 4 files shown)


toddbaert:
@beeme1mr writes Python now :sunglasses:

toddbaert (approved):
Makes sense to me conceptually.

federicobond (approved):
LGTM

federicobond:
Looks like some of the gherkin tests are failing due to the InMemoryFlag change, could you update those so it's ready to merge?

beeme1mr:
@federicobond I'll try 😟

beeme1mr:
It was easier than I thought... perhaps I am a Python developer now 😄

PR #249, opened 2023-12-28
chore: replace black, fake8 and isort with ruff and ruff-format

### LINKED ISSUE #219: Consider migrating from flake8/isort to ruff
[ruff](https://github.com/charliermarsh/ruff) is a very fast Python linter written in Rust. It supports a large amount of rules and is being used by more and more high-profile Python projects.

ruff can replace and expand on the tools we are using right now, such as flake8 and isort.

I propose we migrate our current lint configuration to ruff.

gruebel:
## This PR

- replace `black`, `flake8` and `isort` with `ruff` and `ruff-format`
- enabled the rules, which replace `flake8` and the plugins you mentioned in the `pre-commit` config
- additionally enabled some more, which I used in an other project, but feel free to enable more, if something peaks your interest 🙂 
- adjusted GHA to use the `pre-commit` action instead

### Related Issues

Fixes #219

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-max-line-length = 88
-exclude = .venv/*,venv/*,.git,__pycache__
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,7 +69,7 @@ jobs:
         run: pip install -r requirements.txt
 
       - name: Run pre-commit
-        uses: pre-commit/action@v3
+        uses: pre-commit/action@v3.0.0
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,7 +69,7 @@ jobs:
         run: pip install -r requirements.txt
 
       - name: Run pre-commit
-        run: pre-commit/action@v3
+        uses: pre-commit/action@v3
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v3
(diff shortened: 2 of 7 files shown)


gruebel:
hey @federicobond and @beeme1mr created a new PR, which only does the ruff addition. Will do the code adjustments in a follow-up PR.

beeme1mr (approved):

beeme1mr:
Hey @gruebel, sorry for the delay here. Most of the team was out for the holidays. Things should speed up now that people are coming back 😄

gruebel:
@beeme1mr no problem 😄  Fixed the issue with the missing GHA version tag. It looks like they don't offer a shortened variant for it. I usually use commit hashes to pin all GHA versions. Are you interested in it or you want to keep it as is?

beeme1mr:
> @beeme1mr no problem 😄 Fixed the issue with the missing GHA version tag. It looks like they don't offer a shortened variant for it. I usually use commit hashes to pin all GHA versions. Are you interested in it or you want to keep it as is?

I have a slight preference for version tags because they're easier to understand at a glance but I would be fine with either approach.

gruebel:
> I have a slight preference for version tags because they're easier to understand at a glance but I would be fine with either approach.

All good, then let's keep as it is 🙂

beeme1mr:
I'll give @federicobond more time since he's our resident Python expert 😄

federicobond:
Sorry for the delay! I think this is good to merge. It's hard to understand all the implications of moving to a new set of style rules, but this moves us in the right direction regarding tooling, and if it turns out we need to fine-tune the config we can adjust it later.

federicobond (approved):

PR #255, opened 2024-01-06
ci: add mypy type checking and fix/exclude minor issues

gruebel:
## This PR

- most of the code has type annotation, so this adds a type checking task inside `pre-commit`, which will be triggered during CI
- had to adjust the `StrEnum` import to use the Python system info, so the type checker can properly handle it
- the ignores in `MappingProxyType` are needed, because the original type hints are different

### Follow-up Tasks

- plan is to enable more and more `mypy` checks with follow-up PRs till `strict` mode can be enabled

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,3 +14,9 @@ repos:
       - id: check-yaml
       - id: trailing-whitespace
       - id: check-merge-conflict
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.8.0
+    hooks:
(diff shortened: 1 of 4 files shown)


federicobond (commented):
There is no `openfeature/__init__.py` file to turn `openfeature` into an implicit namespace package. See https://github.com/open-feature/python-sdk/pull/199

Other than that, LGTM.

gruebel:
ok, no problem. Need to adjust the `mypy` config a bit to make it work.

federicobond (approved):

PR #256, opened 2024-01-06
ci: split lint job into lint and sast

gruebel:
## This PR

- typically CodeQL scanning takes longer than other steps, therefore it is better to have a dedicated job for it
- also removed the `Install dependencies` step, because there are no dependencies needed for scanning the project (this could change in the future and then it would make sense to split into 2 different `requirements.txt` files or use something like `poetry` to handle dev and runtime dependencies)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,17 +63,23 @@ jobs:
           python-version: "3.11"
           cache: "pip"
 
+      - name: Run pre-commit
+        uses: pre-commit/action@v3.0.0
+
+  sast:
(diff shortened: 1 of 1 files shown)


federicobond:
What do you think about running CodeQL only for merged commits (on main branch)?

gruebel:
From my experience it is usually too late. If you for example added some suspicious code, then it is merged and flagged, and you will need to either exclude those lines or fix them in a follow-up PR.

beeme1mr (approved):

federicobond (approved):

PR #257, opened 2024-01-08
chore: enable mypy strict mode

gruebel:
## This PR

- most of the type hint changes are pretty straight forward
- one check of the `strict` checks is disabled to keep this PR rather small
- changed a list-filter-lambda combo to a list comprehension, because `mypy` tripped on it and in general list comprehension are more performant compared to such a combo
- also enabled `mypy` pretty mode, which gives quite nice error outputs
ex. 
```python
openfeature/hook/hook_support.py:52: error: Missing type parameters for generic
type "Mapping"  [type-arg]
        hints: typing.Optional[typing.Mapping] = None,
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

### DIFF
--- a/openfeature/_backports/strenum.py
+++ b/openfeature/_backports/strenum.py
@@ -1,7 +1,8 @@
 import sys
 
 if sys.version_info >= (3, 11):
-    from enum import StrEnum
+    # re-export needed for type checking
+    from enum import StrEnum as StrEnum
 else:
     from enum import Enum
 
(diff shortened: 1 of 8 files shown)


federicobond (approved):
Awesome work! LGTM

federicobond:
Apparently one of the lint rules for ruff conflicts with the mypy configuration:
```
openfeature/_backports/strenum.py:5:22: PLC0414 Import alias does not rename original package
```

gruebel:
@federicobond strange I thought I ran `ruff` multiple times 🤔 will ignore the rule in this specific line, because there is no other way I know to make both `mypy` and `ruff` happy. I prefer `mypy` to be happy 😄

PR #261, opened 2024-01-17
feat: improve logging setup

federicobond:
## This PR

Replaces the root logger with a custom logger for all logs emitted by the package. 

I believe this should cover the logging requirements for the SDK, since logs can be manipulated later via handlers and formatters using the standard library `logging` package interface.

### Follow-up Tasks

 - [ ] If we agree, mark this package as supporting logging configuration in the README.


gruebel (approved):
👍

beeme1mr (approved):

PR #264, opened 2024-01-26
fix: Allow string values for `FlagEvaluationDetails.reason` and  `FlagResolutionDetails.reason`

### LINKED ISSUE #262: FlagResolutionDetails reason field should accept any string
The [specification for resolution details](https://github.com/open-feature/spec/blob/main/specification/types.md#resolution-details) states that the reason field should be an optional string. There are some pre-defined reason values (reflected in the `Reason` enum), but a provider should be free to provide an arbitrary reason.

I believe the `FlagResolutionDetails.reason` field should be updated to store an `Optional[str]` instead of the current `Optional[Reason]`.

Happy to work on the PR if this seems acceptable.

See the [related issue on the SDK spec](https://github.com/open-feature/spec/issues/236).

keelerm84:
## This PR

Expands the types allowed for `reason` on `FlagEvaluationDetails` and `FlagResolutionDetails` to include both `Reason` and `str` types.

### Related Issues

Fixes #262

### Notes

Related to spec issue: https://github.com/open-feature/spec/issues/236

### DIFF
--- a/openfeature/flag_evaluation.py
+++ b/openfeature/flag_evaluation.py
@@ -40,7 +40,7 @@ class FlagEvaluationDetails(typing.Generic[T_co]):
     value: T_co
     variant: typing.Optional[str] = None
     flag_metadata: FlagMetadata = field(default_factory=dict)
-    reason: typing.Optional[Reason] = None
+    reason: typing.Optional[typing.Union[str, Reason]] = None
     error_code: typing.Optional[ErrorCode] = None
     error_message: typing.Optional[str] = None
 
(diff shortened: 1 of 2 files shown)


gruebel (approved):
Nice job 🍻

beeme1mr (approved):

lukas-reining (approved):
Not a huge python expert but the change and tests look good!

federicobond (commented):
Hi @keelerm84! Thank you for your contribution. Change LGTM, but I am lost on what the test is supposed to assert.

If it's just a type check then it should be already covered by mypy. Would any of the tests break if you reverted the type change?

keelerm84:
> Hi @keelerm84! Thank you for your contribution. Change LGTM, but I am lost on what the test is supposed to assert.
> 
> If it's just a type check then it should be already covered by mypy. Would any of the tests break if you reverted the type change?

That's a good point. I've removed the useless test and will rely on mypy as suggested.

PR #266, opened 2024-01-28
feat: make specific fields in HookContext immutable

gruebel:
## This PR

- also added tests to cover section 4.1

### Related Issues

Relates #2

### DIFF
--- a/openfeature/hook/__init__.py
+++ b/openfeature/hook/__init__.py
@@ -29,6 +29,11 @@ class HookContext:
     client_metadata: typing.Optional[ClientMetadata] = None
     provider_metadata: typing.Optional[Metadata] = None
 
+    def __setattr__(self, key: str, value: typing.Any) -> None:
+        if hasattr(self, key) and key in ("flag_key", "flag_type", "default_value"):
+            raise AttributeError(f"Attribute {key!r} is immutable")
+        super().__setattr__(key, value)
+
 
 class Hook:
     def before(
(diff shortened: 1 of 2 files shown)


openfeature/hook/__init__.py
```diff
@@ -29,6 +29,11 @@ class HookContext:
     client_metadata: typing.Optional[ClientMetadata] = None
     provider_metadata: typing.Optional[Metadata] = None
 
+    def __setattr__(self, key: str, value: typing.Any) -> None:
+        if hasattr(self, key) and key in ("flag_key", "flag_type", "default_value"):
```
federicobond, line 33: Do we need the `hasattr(self, key)` check? Or can we do with just the `in` expression?

openfeature/hook/__init__.py
```diff
@@ -29,6 +29,11 @@ class HookContext:
...
+            raise AttributeError(f"Attribute {key!r} is immutable")
```
federicobond, line 34: What's the difference with just `{key}` here?

gruebel: it just adds quotes around it 😄 ex. `Attribute 'flag_key' is immutable`

openfeature/hook/__init__.py
gruebel: yeah, I tried it without it first, but then it fails on the initialization of the dataclass instance.

federicobond: ah makes sense

federicobond (approved):

PR #268, opened 2024-02-02
feat: add FeatureProvider protocol

federicobond:
This PR introduces a new FeatureProvider protocol. Public API methods can use it as an argument type when they want to denote any provider, instead of AbstractProvider which is a bit clunky.

AbstractProvider remains as a provider base class for backward compatibility but it's import path at `openfeature.provider.provider` does not have the best ergonomics and we should begin working on a migration path before 1.0 is released.

### DIFF
--- a/openfeature/provider/__init__.py
+++ b/openfeature/provider/__init__.py
@@ -7,7 +7,7 @@ from openfeature.hook import Hook
 from .metadata import Metadata
 
 
-class FeatureProvider(typing.Protocol):
+class FeatureProvider(typing.Protocol):  # pragma: no cover
     def initialize(self, evaluation_context: EvaluationContext) -> None:
         ...
 
--- a/openfeature/provider/__init__.py
+++ b/openfeature/provider/__init__.py
@@ -1,0 +1,61 @@
+import typing
+
+from openfeature.evaluation_context import EvaluationContext
+from openfeature.flag_evaluation import FlagResolutionDetails
+from openfeature.hook import Hook
+
+from .metadata import Metadata
+
+
+class FeatureProvider(typing.Protocol):
+    def initialize(self, evaluation_context: EvaluationContext) -> None:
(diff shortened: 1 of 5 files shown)


beeme1mr (approved):

openfeature/provider/__init__.py
```diff
@@ -0,0 +1,61 @@
+import typing
+
+from openfeature.evaluation_context import EvaluationContext
+from openfeature.flag_evaluation import FlagResolutionDetails
+from openfeature.hook import Hook
+
+from .metadata import Metadata
+
+
+class FeatureProvider(typing.Protocol):
```
gruebel, line 10:
```suggestion
class FeatureProvider(typing.Protocol):  # pragma: no cover
```
I think, this should disable coverage checking for the whole class, which we anyway don't need to do here

gruebel (approved):
always love to see more usage of `Protocol` in Python projects

openfeature/provider/__init__.py
federicobond: Interesting suggestion! I usually use coverage tools to point me to what code has insufficient testing but never as a target itself. What is the policy in other OpenFeature projects?  Should we aim for 100% coverage here?

federicobond: We are already using `pragma: no cover` in some places, so it's not really a big deal

toddbaert: > Interesting suggestion! I usually use coverage tools to point me to what code has insufficient testing but never as a target itself. What is the policy in other OpenFeature projects? Should we aim for 100% coverage here?

It's just a metric. We aim for high coverage, but coverage isn't always accurate. Some small PRs will have good testing but decrease overall coverage by 0.1% or something... As a maintainer you can bypass it. I recommend just using your judgement and not making it a blocker.

Issue #125, opened 2023-06-28
Changes to support spec `v0.6.0` compliance


toddbaert:
:warning: 0.6.0 should probably be undertaken only after the existing [1.0-functional requirements](https://github.com/open-feature/python-sdk/issues?q=is%3Aissue+is%3Aopen+label%3A1.0-release) are complete.

## Implement specification v0.6.0 changes

See [changes since last release](https://github.com/open-feature/spec/compare/v0.5.2...v0.6.0).

Please note there's no need to complete all these tasks in a single PR (in fact, that's not recommended). Consider creating a new sub issue for each of the below requirements.

### Functional requirements:

- [x] implement flag metadata: https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a
#212
- [x] implement named client support
- [x] Implement Initialize/Shutdown on provider registration: https://github.com/open-feature/spec/commit/a4ffec3d4416b8ad6d12d1ebfa2faf81ad153f65
#213 
- [x] Implement events https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md
  - [x] #131 (API level events)

### Non-functional requirements:

- [x] add relevant tests
- [x] add examples and documentation, see template README
- [x] update specification badge to `v0.6.0`: [![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.6.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.6.0)

### Keep in mind:

- backwards compatibility is _highly encouraged_
  - take every precaution to ensure existing code written by _application authors_ and existing providers implementations, do not break
- it's recommended these features are released together

federicobond:
There are no more directly actionable items in this issue 🎉 , so we are ready to close and handle any follow-up tasks in separate issues.

PR #271, opened 2024-02-03
feat!: add support for domains

federicobond:
## This PR

Adds support for domains with a minimalist implementation. Feedback is especially welcome around public APIs and backward compatibility issues.

Builds on top of https://github.com/open-feature/python-sdk/pull/270

### Related Issues

Refs https://github.com/open-feature/python-sdk/issues/125

### Follow-up Tasks
  - [ ] Ensure the API is idiomatic and matches the spec guidelines
  - [ ] Test backward compatibility with existing clients and providers

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ print("Value: " + str(flag_value))
 | ✅      | [Targeting](#targeting)         | Contextually-aware flag evaluation using [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context). |
 | ✅      | [Hooks](#hooks)                 | Add functionality to various stages of the flag evaluation life-cycle.                                                             |
 | ✅      | [Logging](#logging)             | Integrate with popular logging packages.                                                                                           |
-| ❌      | [Named clients](#named-clients) | Utilize multiple providers in a single application.                                                                                |
+| ✅      | [Domains](#domains)             | Logically bind clients with providers application.                                                                                 |
 | ❌      | [Eventing](#eventing)           | React to state changes in the provider or flag management system.                                                                  |
 | ✅      | [Shutdown](#shutdown)           | Gracefully clean up a provider during application shutdown.                                                                        |
 | ✅      | [Extending](#extending)         | Extend OpenFeature with custom providers and hooks.                                                                                |
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ print("Value: " + str(flag_value))
 | ✅      | [Targeting](#targeting)         | Contextually-aware flag evaluation using [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context). |
 | ✅      | [Hooks](#hooks)                 | Add functionality to various stages of the flag evaluation life-cycle.                                                             |
 | ✅      | [Logging](#logging)             | Integrate with popular logging packages.                                                                                           |
-| ✅      | [Domains](#domains)             | Logically bind clients with providers application.                                                                                 |
+| ✅      | [Domains](#domains)             | Logically bind clients with providers.                                                                                             |
 | ❌      | [Eventing](#eventing)           | React to state changes in the provider or flag management system.                                                                  |
 | ✅      | [Shutdown](#shutdown)           | Gracefully clean up a provider during application shutdown.                                                                        |
 | ✅      | [Extending](#extending)         | Extend OpenFeature with custom providers and hooks.                                                                                |
@@ -128,8 +128,8 @@ api.set_provider(NoOpProvider())
 open_feature_client = api.get_client()
 ```
 
-<!-- In some situations, it may be beneficial to register multiple providers in the same application.
-This is possible using [named clients](#named-clients), which is covered in more detail below. -->
+In some situations, it may be beneficial to register multiple providers in the same application.
+This is possible using [domains](#domains), which is covered in more detail below.
 
 ### Targeting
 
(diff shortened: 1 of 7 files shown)


openfeature/api.py
```diff
@@ -14,31 +14,52 @@
...
     provider.initialize(_evaluation_context)
 
 
-def get_provider() -> typing.Optional[FeatureProvider]:
-    global _provider
-    return _provider
+def _set_domain_provider(domain: str, provider: FeatureProvider) -> None:
+    if domain in _providers:
+        old_provider = _providers[domain]
+        del _providers[domain]
```
gruebel, line 46: if you want to use Python 3.8 syntax
```suggestion
    if old_provider := _providers.pop(domain, None):
```

openfeature/api.py
```diff
@@ -14,31 +14,52 @@
 
 _hooks: typing.List[Hook] = []
 
+_providers: typing.Dict[str, FeatureProvider] = {}
+
 
 def get_client(
-    name: typing.Optional[str] = None, version: typing.Optional[str] = None
+    domain: typing.Optional[str] = None, version: typing.Optional[str] = None
```
gruebel, line 21: this is basically a breaking change for anyone invoking it with keyword arguments.

openfeature/client.py
```diff
@@ -61,26 +60,29 @@
...
-    name: typing.Optional[str]
+    name: typing.Optional[str] = None
+    domain: typing.Optional[str] = None
 
 
 class OpenFeatureClient:
     def __init__(
         self,
-        name: typing.Optional[str],
+        domain: typing.Optional[str],
```
gruebel, line 70: same here

openfeature/client.py
```diff
@@ -61,26 +60,29 @@
...
-        self.name = name
+        self.domain = domain
         self.version = version
         self.context = context or EvaluationContext()
         self.hooks = hooks or []
-        self.provider = provider
+
+    @property
+    def provider(self) -> FeatureProvider:
+        return api.get_provider(domain=self.domain)
```
gruebel, line 82: you probably need to also implement `@provider.setter` otherwise you wouldn't be able to change it via the attribute anymore, except it is intended.

gruebel (commented):
Due to the argument renaming it becomes a breaking change and then you actually could completely get rid of `_provider` and initialize `_providers` like this
```python
_providers: typing.Dict[typing.Optional[str], AbstractProvider] = {None: NoOpProvider()}
```

openfeature/api.py
federicobond: It's is. Are we ok with pushing this breaking change? Or should we have a deprecation period?

gruebel: Im totally fine with breaking changes, we just increase the minor version instead.

openfeature/api.py
```diff
@@ -14,31 +14,52 @@
...
+        del _providers[domain]
+        if old_provider not in _providers.values():
+            old_provider.shutdown()
+    if provider not in _providers.values():
+        provider.initialize(_evaluation_context)
+    _providers[domain] = provider
 
 
-def get_provider_metadata() -> typing.Optional[Metadata]:
+def get_provider(domain: typing.Optional[str] = None) -> FeatureProvider:
```
beeme1mr, line 54: This should be a private method.

openfeature/api.py
beeme1mr, line 1: You may want to consider adding a `clear_providers` method that removes all providers and resets the default provider to NoOp. This makes writing tests easier.

openfeature/api.py
toddbaert: Since this is a sub 1.0 SDK, I'd recommend just changing the title to `feat!...` (note the !) which will mark this as a breaking change in release-please. I think breaking changes are fine sub 1.0 and I don't think it's worth the maintenance burden to do non-breaking deprecations (just my 2c).

openfeature/client.py
```diff
@@ -61,26 +60,29 @@
...
-        self.name = name
+        self.domain = domain
         self.version = version
         self.context = context or EvaluationContext()
         self.hooks = hooks or []
-        self.provider = provider
+
+    @property
+    def provider(self) -> FeatureProvider:
+        return api._provider_registry.get_provider(self.domain)
```
federicobond, line 82: This is not super clean. Should we inject the registry in the client constructor?

federicobond:
I introduced a ProviderRegistry in the last commit similar to the ones in other SDKs because the code was becoming a bit convoluted.

README.md
```diff
@@ -105,7 +105,7 @@ print("Value: " + str(flag_value))
 | ✅      | [Targeting](#targeting)         | Contextually-aware flag evaluation using [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context). |
 | ✅      | [Hooks](#hooks)                 | Add functionality to various stages of the flag evaluation life-cycle.                                                             |
 | ❌      | [Logging](#logging)             | Integrate with popular logging packages.                                                                                           |
-| ❌      | [Named clients](#named-clients) | Utilize multiple providers in a single application.                                                                                |
+| ✅      | [Domains](#domains)             | Logically bind clients with providers application.                                                                                 |
```
beeme1mr, line 108:
```suggestion
| ✅      | [Domains](#domains)             | Logically bind clients with providers.                                                                                 |
```

openfeature/client.py
beeme1mr: That's what we do in the JS SDK.

openfeature/provider/registry.py
```diff
@@ -0,0 +1,59 @@
...
+            raise GeneralError(error_message="No provider")
+        if self._default_provider:
+            self._default_provider.shutdown()
+        self._default_provider = provider
+        provider.initialize(self._get_evaluation_context())
+
+    def get_default_provider(self) -> FeatureProvider:
+        return self._default_provider
+
+    def clear_providers(self) -> None:
```
beeme1mr, line 46: Is it possible to clear or reset the default provider?

beeme1mr (approved):

openfeature/provider/registry.py
federicobond: It should clear to NoOpProvider, right?

beeme1mr: Yeah, exactly.

openfeature/api.py
federicobond: I like this, but the linter forced me to put the nested if in the same line and it was not as readable as the current version.

federicobond:
Rebased on top of latest `main`.

README.md
beeme1mr, line 1: Please update the code examples and the **named client** section in the readme to use domains.

toddbaert (approved):

PR #278, opened 2024-02-21
feat: implement provider events

federicobond:
## This PR

- Adds support for eventing according to the spec.

### Follow-up Tasks

 - [ ] Improve tests


openfeature/api.py
```diff
@@ -67,3 +81,30 @@ def get_hooks() -> typing.List[Hook]:
...
+) -> None:
+    details = EventDetails.from_provider_event_details(
+        provider.get_metadata().name, provider_details
+    )
+    # run the global handlers
+    _event_support.run_global_handlers(event, details)
+    # run the handlers for clients associated to this provider
+    for client in _clients_with_handlers:
+        if client.provider == provider:
+            _event_support.run_client_handlers(client, event, details)
```
federicobond, line 110: This is the key dispatching logic to review, not sure I got it right, might need further tweaking.

README.md
```diff
@@ -214,7 +214,26 @@ For more details, please refer to the [providers](#providers) section.
...
+Please refer to the documentation of the provider you're using to see what events are supported.
+
+```python
+from openfeature import api
+from openfeature.provider import ProviderEvent
+
+def on_provider_ready(event_details: EventDetails):
+    print(f"Provider {event_details.provider_name} is ready")
+
+api.on(ProviderEvent.PROVIDER_READY, on_provider_ready)
```
federicobond, line 228: We could use `add_handler` instead of `on` here. Don't have a particularly strong preference for either.

toddbaert:
I'll take a close look at this in the next couple days.

openfeature/event.py
```diff
@@ -0,0 +1,100 @@
...
+            metadata=details.metadata,
+        )
+
+
+EventHandler = Callable[[EventDetails], None]
+
+
+class EventSupport:
+    _global_handlers: Dict[ProviderEvent, List[EventHandler]]
+    _client_handlers: Dict["OpenFeatureClient", Dict[ProviderEvent, List[EventHandler]]]
```
toddbaert, line 50: It's probably due to my inexperience with Python, but I don't understand the use of this "OpenFeatureClient" key.

openfeature/event.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def run_global_handlers(self, event: ProviderEvent, details: EventDetails) -> None:
+        for handler in self._global_handlers[event]:
+            handler(details)
+
+    def add_client_handler(
+        self, client: "OpenFeatureClient", event: ProviderEvent, handler: EventHandler
+    ) -> None:
+        handlers = self._client_handlers[client][event]
+        handlers.append(handler)
```
toddbaert, line 70: I think we're missing and implementation of [5.3.3](https://openfeature.dev/specification/sections/events/#requirement-533)

tests/test_client.py
```diff
@@ -182,3 +183,56 @@ def test_should_call_api_level_hooks(no_op_provider_client):
...
+    spy = MagicMock()
+
+    client = get_client()
+    client.add_handler(ProviderEvent.PROVIDER_READY, spy.provider_ready)
+    client.remove_handler(ProviderEvent.PROVIDER_READY, spy.provider_ready)
+
+    provider_details = ProviderEventDetails(message="message")
+    provider.emit_provider_ready(provider_details)
+
+    spy.provider_ready.assert_not_called()
```
toddbaert, line 238: I think we need some tests to assert that bound handlers support the late binding of providers. When event handlers are added to `clientX`, and `clientX` is associated with `providerX`, they should run if `providerX` emits events. If `clientX` is then re-associated with `providerY`, then `clientX`'s handlers should run only if `providerY` emits events. Note that this applies to the default provider as well (`providerX` might be the default provider).

openfeature/event.py
gruebel: It just tells the type checker to expect a  `OpenFeatureClient`  class or subclass as the key in the dict.

openfeature/provider/provider.py
```diff
@@ -66,3 +67,22 @@ def resolve_object_details(
         evaluation_context: typing.Optional[EvaluationContext] = None,
     ) -> FlagResolutionDetails[typing.Union[dict, list]]:
         pass
+
+    def emit_provider_ready(self, details: ProviderEventDetails) -> None:
```
toddbaert, line 71: The ability to manually emit this and other events is needed.

In addition, the SDK should automatically emit READY/ERROR based on the suceess/failure of the init function.

[5.3.1](https://openfeature.dev/specification/sections/events/#requirement-531), [5.3.2](https://openfeature.dev/specification/sections/events/#requirement-532). Again there's a bit more info in the unreleased specification: https://github.com/open-feature/spec/blob/31da456d953752522c557ee0b7cce7db7bba331e/specification/sections/05-events.md?plain=1#L130

openfeature/event.py
```diff
@@ -0,0 +1,100 @@
...
+
+@dataclass
+class ProviderEventDetails:
+    flags_changed: Optional[List[str]] = None
+    message: Optional[str] = None
+    metadata: Dict[str, Union[bool, str, int, float]] = field(default_factory=dict)
+
+
+@dataclass
+class EventDetails(ProviderEventDetails):
```
toddbaert, line 27: In the unreleased spec version, there's also a optional (nullable) [errorCode](https://github.com/open-feature/python-sdk/blob/141858d2359bf6bf439426b3ea4ba322f4b10421/openfeature/exception.py#L5) here: https://github.com/open-feature/spec/blob/main/specification/types.md#event-details

I suggest you add it now.

toddbaert (dismissed):
Please pardon my lack of Python skills if I'm missed something.

Overall the structures and API look great!

I left some requested changes specifically on the finer points of the eventing spec. They can be tricky, and some of it could be better fleshed-out in the spec, as I think you noted before. I will work on some additional supporting explanations today. Thanks for your efforts here! :pray:

openfeature/event.py
toddbaert: Thanks!

federicobond: Yes, it's Python's clunky way of dealing with circular module imports or forward reference problems.

The annotation is understood by mypy which looks up the name in the scope, and works together with the `if TYPE_CHECKING` clause above. That constant is only True during mypy's static analysis, which has no problems with circular imports.

openfeature/provider/provider.py
federicobond: Been thinking about how to approach these requirements and I think it's best to first implement provider status in a separate pull request and then integrate those changes here.

openfeature/event.py
federicobond: Done!

federicobond:
This should be ready to review from a specification perspective, we might want to add some further tests maybe.

openfeature/event.py
```diff
@@ -0,0 +1,131 @@
...
+    _client_handlers: Dict[OpenFeatureClient, Dict[ProviderEvent, List[EventHandler]]]
+
+    def __init__(self) -> None:
+        self._global_handlers = defaultdict(list)
+        self._client_handlers = defaultdict(lambda: defaultdict(list))
+
+    def run_client_handlers(
+        self, client: OpenFeatureClient, event: ProviderEvent, details: EventDetails
+    ) -> None:
+        for handler in self._client_handlers[client][event]:
```
matthewelwell, line 74: I don't fully understand the internals of python's hashing so maybe this isn't an issue, but it surprised me to see the client instance being used as the key here rather than some unique attribute on the client (although I'm not sure there is one!)

openfeature/event.py
```diff
@@ -0,0 +1,131 @@
...
+        provider_details: ProviderEventDetails,
+    ) -> None:
+        details = EventDetails.from_provider_event_details(
+            provider.get_metadata().name, provider_details
+        )
+        # run the global handlers
+        self.run_global_handlers(event, details)
+        # run the handlers for clients associated to this provider
+        for client in self._client_handlers:
+            if client.provider == provider:
```
matthewelwell, line 120: I see here that my comment above regarding the dictionary key is relevant here too. I guess without using the client itself, we'd have to either include it in the dictionary value, or create some kind of registry.

openfeature/provider/provider.py
```diff
@@ -66,3 +67,22 @@ def resolve_object_details(
...
+    def emit_provider_error(self, details: ProviderEventDetails) -> None:
+        self.emit(ProviderEvent.PROVIDER_ERROR, details)
+
+    def emit_provider_stale(self, details: ProviderEventDetails) -> None:
+        self.emit(ProviderEvent.PROVIDER_STALE, details)
+
+    def emit(self, event: ProviderEvent, details: ProviderEventDetails) -> None:
+        from openfeature.api import _run_handlers_for_provider
+
+        _run_handlers_for_provider(self, event, details)
```
matthewelwell, line 88: Should this function be public?

openfeature/provider/registry.py
```diff
@@ -1,6 +1,10 @@
 import typing
 
 from openfeature.evaluation_context import EvaluationContext
+from openfeature.event import (
+    ProviderEventDetails,
+    _provider_status_to_event,
```
matthewelwell, line 6: Should this function also be made public?

matthewelwell (commented):
@federicobond I've added a few comments here from a python perspective but I don't purport to be an expert so feel free to ignore as you see fit!

openfeature/client.py
```diff
@@ -439,6 +440,12 @@ def _create_provider_evaluation(
             error_message=resolution.error_message,
         )
 
+    def add_handler(self, event: ProviderEvent, handler: EventHandler) -> None:
+        api._add_client_handler(self, event, handler)
```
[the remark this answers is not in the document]
federicobond: Yes, they are convenience methods for the internal implementation only, end users should use `OpenFeatureClient.add_handler` and `OpenFeatureClient.remove_handler`.

I think adding them to `openfeature.api` though makes things a bit confusing, I'll try to clean that up.

openfeature/event.py
federicobond: Client instances hash to object identity, basically only the exact same instance in memory will match. I'll look closer into this, but apparently, it should conform to spec. Perhaps registering at the domain works just as well?

openfeature/provider/provider.py
federicobond: Not sure why? I wouldn't want end users calling it directly.

openfeature/provider/registry.py
federicobond: We could put it as a ProviderStatus method perhaps?

federicobond:
From Requirement 5.1.1:

> Providers cannot emit PROVIDER_CONTEXT_CHANGED or PROVIDER_RECONCILING event.

Should we add a guard to make sure this cannot happen?

beeme1mr (approved):

openfeature/provider/registry.py
matthewelwell: Yeah, I think a classmethod on the `ProviderStatus` class would be ideal here!

matthewelwell (commented):
Apologies for the delay in responding here @federicobond, I really need to look into my GH notifications 🙄 

I've added a few more comments in response to yours. Again, I'm happy to be ignored on any of them if you see fit.

openfeature/event.py
federicobond: Ah, performance should be negligible because object hashes default to using the memory address of the instance as key, which is super fast to compute.

openfeature/provider/registry.py
federicobond: I ended up adding a classmethod to `ProviderEvent` because I think it makes more sense to place the dependency in that direction, happy to reconsider though.

federicobond:
@matthewelwell I refactored the EventSupport implementation so that we are no longer using private module imports. It should be much cleaner now.

openfeature/_event_support.py
```diff
@@ -0,0 +1,89 @@
...
+def run_client_handlers(
+    client: OpenFeatureClient, event: ProviderEvent, details: EventDetails
+) -> None:
+    for handler in _client_handlers[client][event]:
+        handler(details)
+
+
+def run_global_handlers(event: ProviderEvent, details: EventDetails) -> None:
+    for handler in _global_handlers[event]:
+        handler(details)
```
gruebel, line 33: depends a bit on the specs, but this will fail with an ugly `KeyError`, if someone invokes it with a non-existing or not yet added event/client. Wouldn't it be better just to log an error message and not through an exception?

openfeature/_event_support.py
```diff
@@ -0,0 +1,89 @@
...
+    handlers.append(handler)
+
+    _run_immediate_handler(client, event, handler)
+
+
+def remove_client_handler(
+    client: OpenFeatureClient, event: ProviderEvent, handler: EventHandler
+) -> None:
+    handlers = _client_handlers[client][event]
+    handlers.remove(handler)
```
gruebel, line 49: do we want to be more forgiving when something is tried to be removed, but doesn't exist anymore?

openfeature/_event_support.py
```diff
@@ -0,0 +1,89 @@
...
+def add_global_handler(event: ProviderEvent, handler: EventHandler) -> None:
+    _global_handlers[event].append(handler)
+
+    from openfeature.api import get_client
+
+    _run_immediate_handler(get_client(), event, handler)
+
+
+def remove_global_handler(event: ProviderEvent, handler: EventHandler) -> None:
+    _global_handlers[event].remove(handler)
```
gruebel, line 61: same here

openfeature/event.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class ProviderEvent(Enum):
+    PROVIDER_READY = "PROVIDER_READY"
+    PROVIDER_CONFIGURATION_CHANGED = "PROVIDER_CONFIGURATION_CHANGED"
+    PROVIDER_ERROR = "PROVIDER_ERROR"
+    PROVIDER_FATAL = "PROVIDER_FATAL"
+    PROVIDER_STALE = "PROVIDER_STALE"
+
+    __status__: ClassVar[Dict[ProviderStatus, str]] = {
```
gruebel, line 18: nice, rare to see the usage of `typing. ClassVar` 🍸

openfeature/provider/provider.py
```diff
@@ -66,3 +68,20 @@ def resolve_object_details(
...
+    ) -> None:
+        self.emit(ProviderEvent.PROVIDER_CONFIGURATION_CHANGED, details)
+
+    def emit_provider_error(self, details: ProviderEventDetails) -> None:
+        self.emit(ProviderEvent.PROVIDER_ERROR, details)
+
+    def emit_provider_stale(self, details: ProviderEventDetails) -> None:
+        self.emit(ProviderEvent.PROVIDER_STALE, details)
+
+    def emit(self, event: ProviderEvent, details: ProviderEventDetails) -> None:
```
gruebel, line 86: should this be a "private" method or is it ok to expose it to the user?

gruebel (commented):
great work, just few minor comments/questions

openfeature/_event_support.py
federicobond: `_global_handlers` is a `defaultdict(list)` so it should just return an empty list for non-existent events.

openfeature/provider/provider.py
federicobond: I don't think it hurts to leave it, but it does introduce the possibility of emitting prohibited events. We can take a cue from other SDKs and check what they do.

openfeature/_event_support.py
gruebel: yeah, you are right 🙈

openfeature/provider/provider.py
gruebel: good point, in the JS SDK it is public 😄

gruebel (approved):

matthewelwell (approved):
Looks good, thanks @federicobond !

federicobond:
Merged! 🎉  Thanks everyone for your helpful comments and suggestions.

beeme1mr:
Amazing, nice job @federicobond!

PR #285, opened 2024-03-02
refactor: improve Hook Hints typing

gruebel:
## This PR

- it doesn't really enforce `4.2.2.1`, but a type checker will complain, if you try to change a value in the hook hints
- had to switch from `dataclass` to a normal class `HookContext` to make the metadata fields immutable


federicobond (approved):
LGTM!

PR #288, opened 2024-03-04
feat: implement provider status

federicobond:
This self-contained change introduces the functionality required to properly implement provider status events later. It has no user-visible changes at this point.

I wanted to split this out of the events pull request because it had the potential to snowball the size of the diff and lower the quality of the review in general.


openfeature/provider/registry.py
```diff
@@ -50,10 +53,31 @@ def clear_providers(self) -> None:
...
         from openfeature.api import get_evaluation_context
 
         return get_evaluation_context()
+
+    def _initialize_provider(self, provider: FeatureProvider) -> None:
+        try:
+            if hasattr(provider, "initialize"):
+                provider.initialize(self._get_evaluation_context())
+            self._set_provider_status(provider, ProviderStatus.READY)
+        except Exception:
```
toddbaert, line 69: It should be possible for the provider to signal that the thrown exception is not recoverable (FATAL): https://github.com/open-feature/spec/blob/37cf68b0d68b6814514bcded521b9e199efcead3/specification/sections/02-providers.md?plain=1#L195

Here's something similar in JS: https://github.com/open-feature/js-sdk/blob/97a34b630eb7caa7ad744034ae56adf0f16e5005/packages/shared/src/open-feature.ts#L282-L288

Doesn't need to be done now but it seems relevant.

openfeature/provider/__init__.py
```diff
@@ -7,6 +8,14 @@
 from .metadata import Metadata
 
 
+class ProviderStatus(Enum):
+    NOT_READY = "NOT_READY"
+    READY = "READY"
+    ERROR = "ERROR"
+    STALE = "STALE"
+    FATAL = "FATAL"
```
toddbaert, line 16: Consider adding the `FATAL` error code as well, relevant to my other comment [here](https://github.com/open-feature/python-sdk/pull/288/files#r1515105556).

toddbaert:
The "short circuiting" behavior mentioned in 1.7.6 and 1.7.7 might be relevant enough to implement here as well: https://github.com/open-feature/spec/blob/37cf68b0d68b6814514bcded521b9e199efcead3/specification/sections/01-flag-evaluation.md?plain=1#L485-L502

toddbaert (approved):
Nothing looks wrong here to me, but there's a few related things that might be worth adding.

toddbaert (approved):
This looks fully spec compliant to me, as far as provider status stuff goes.

gruebel (approved):
nice work 🏅

PR #297, opened 2024-03-20
ci: switch to hatch

### LINKED ISSUE #59: Implement wrapper for packaging and dependency management
Implement an OTS python wrapper for packaging and dependency management (e.g. [Pipenv](https://pipenv.pypa.io/en/latest/index.html) or [Poetry](https://github.com/python-poetry/poetry))

### LINKED ISSUE #133: Use hatch for packaging and dependency management
Poetry helps a lot for packaging and dependency management, it would be great to use it instead of pip.

gruebel:
## This PR

- I stayed close to the PR open-feature/python-sdk-contrib#35

Fixes #59
Fixes #133


federicobond:
Nice work! Let's derisk the current release by waiting to merge until after we release event support 🙏

federicobond (approved):

federicobond:
Sorry for the delay! Looks good to me. Only wondering if we should pin the test dependencies and let renovate handle updates.

gruebel:
> Sorry for the delay! Looks good to me. Only wondering if we should pin the test dependencies and let renovate handle updates.

I have no problem with pinning them, but I couldn't find any info about `renovate` supporting `hatch` envs, therefore I'm hesitant to really pin them.

federicobond:
Ah makes sense, I only thought about the dependencies being in a `pyproject.toml` file, but renovate needs to support env dependencies explicitly.

Issue #214, opened 2023-10-18
Use __all__ in every public module to control which names are exported


federicobond:

gruebel:
Question regarding the expectations here. Do you just want to expose all the current files/modules as is or hide in which file they actually reside.

Ex.
```python
from openfeature import api
from openfeature.client import OpenFeatureClient
from openfeature.provider.in_memory_provider import InMemoryFlag, InMemoryProvider
``` 
or this
```python
from openfeature import api
from openfeature import OpenFeatureClient
from openfeature.provider import InMemoryFlag, InMemoryProvider
```

federicobond:
There is room to keep improving the public module hierarchy. The only thing we should avoid is making multiple releases with breaking changes. If we are going to make changes on this then we should bundle all of them together in a single release so that users only have to migrate once.

My current thinking is that we should use `__all__` to control visibility *within* public modules, to avoid re-exporting imports and such; and we should use underscore-prefixed modules for hiding internal structure, either using `_internal` modules or splitting files into multiple internal prefixed modules that get imported by a single public one.

For a good reference on this check out how the [opentelemetry-python](https://github.com/open-telemetry/opentelemetry-python) SDK does it.

As for your specific question, I don't have a strong preference either way for the OpenFeatureClient import, but the InMemoryProvider one is probably more idiomatic as `openfeature.provider.in_memory`. It makes sense to separate providers into multiple modules to avoid parsing/loading unused providers.

federicobond:
What I have said is not established convention, just my own curated best practices from years of working with Python. Feel free to suggest alternatives.

federicobond:
One big issue with the current module hierarchy is that the API is commonly imported as:

```python
from openfeature import api
```

Or worse:
```python
from openfeature.api import set_provider
```

Neither of these names (`api`, `set_provider`) give any indication as to what they belong to, if you find them in the middle of a long file. A more pythonic approach would be to be able to import them like this:

```python
import openfeature

openfeature.set_provider(provider)
```
Or, if you want:
```python
import openfeature as of

of.set_provider()
```

gruebel:
Typically I prefer to expose the most important ones directly at the root level of the package and less frequent used ones through the their sub namespace. But I think your suggestion can't be implemented, because `openfeature` is a namespace package and therefore you can't set `__all__` in the `__init__.py`. For the sub namespaces it is no problem, but the root won't work sadly.

federicobond:
[...]

PR #306, opened 2024-03-27
fix!: restrict exported names with __all__

federicobond:
Refs https://github.com/open-feature/python-sdk/issues/214

Note: this is a breaking change. We should do an ecosystem check to make sure most packages are using the correct imports.

### DIFF
--- a/openfeature/provider/__init__.py
+++ b/openfeature/provider/__init__.py
@@ -7,6 +7,8 @@ from openfeature.hook import Hook
 
 from .metadata import Metadata
 
+__all__ = ["ProviderStatus", "FeatureProvider", "Metadata"]
+
 
 class ProviderStatus(Enum):
     NOT_READY = "NOT_READY"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,8 @@ jobs:
       - name: Run E2E tests with behave
         run: hatch run e2e
 
-      - name: Upload coverage to Codecov
+      - if: matrix.python-version == '3.11'
+        name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
           flags: unittests # optional
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
         with:
           flags: unittests # optional
           name: coverage # optional
-          fail_ci_if_error: true # optional (default = false)
+          fail_ci_if_error: false # optional (default = false)
           verbose: true # optional (default = false)
 
   lint:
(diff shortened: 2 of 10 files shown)


federicobond:
Should `AbstractProvider` be importable from `openfeature.provider` so we can phase out `openfeature.provider.provider`?

openfeature/provider/__init__.py
```diff
@@ -7,6 +7,8 @@
 
 from .metadata import Metadata
 
+__all__ = ["ProviderStatus", "FeatureProvider", "Metadata"]
```
gruebel, line 10: yeah, let's add here `AbstractProvider` and do we want to expose `ProviderRegistry` too or keep it as an internal component?

gruebel (commented):
nice work 💪

openfeature/provider/__init__.py
federicobond: Yes, I would expose `AbstractProvider` here, but I'd keep `ProviderRegistry` internal unless there is a use case for exposing it.

gruebel:
I think exposing `AbstractProvider` should be done with your other PR #309 then the circular dependency will be gone.

federicobond:
Something weird with the Codecov uploader, @toddbaert @beeme1mr do you know what might be going on?

beeme1mr:
I think this happens when Codecovs attempts to call the GitHub APIs and are throttled. I just ran one of the failed tests to see if that fixed it. If that doesn't help, I will investigate tomorrow.

beeme1mr:
Unfortunately, it looks like a very common issue. 
https://github.com/codecov/codecov-action/issues/598

beeme1mr:
We should probably only run Codecov once after the python specific tests finish.

gruebel:
Yeah, there is no need to upload the results for each Python version, just one is enough. We can add an if condition to only upload for Python 3.11 as it is the default version for dinner other jobs in the same workflow.

gruebel:
ok, seems like we are not the only ones with the codecov issue https://github.com/codecov/codecov-action/issues/1359 seems like we need to upgrade to v4 with API token.

beeme1mr:
> ok, seems like we are not the only ones with the codecov issue [codecov/codecov-action#1359](https://github.com/codecov/codecov-action/issues/1359) seems like we need to upgrade to v4 with API token.

Unfortunately, that has [challenges too](https://github.com/codecov/feedback/issues/112). Perhaps should should consider replacing CodeCov or skipping the upload step for now.

gruebel:
@beeme1mr an alternative could be https://coveralls.io/ thoughts?

federicobond:
I inclined to not rush with the decision and just ignore upload failures in CI temporarily.

beeme1mr (approved):

PR #310, opened 2024-03-28
chore: update renovate config

federicobond:
* Replaces old `pip_requirements` manager which will not match any files with `pep621` which matches pyproject.toml
 * Configures `pre-commit` manager to update pre-commit hooks.

I'm not an expert in Renovate so not sure if this will work but it sounds reasonable.


gruebel (approved):
🥇

PR #313, opened 2024-04-09
chore: update codecov and use OIDC

gruebel:


gruebel:
ok, that won't work

beeme1mr:
@gruebel do you know why that won't work? Using OIDC would be nice.

gruebel:
@beeme1mr Yeah, basically the GH token can't have the needed permission, when coming from a fork. It would only work for PRs from the same repo.

beeme1mr:
Okay, yeah. That was the issue Todd reported a few months ago. CodeCov doesn't seem to have a great answer for that unfortunately.

PR #314, opened 2024-04-10
refactor!: move AbstractProvider to openfeature.provider

federicobond:
Moves `AbstractProvider` to `openfeature.provider` module.

Had to reverse the dependency direction between ProviderStatus and ProviderEvent to avoid a circular dependency.

This release SHOULD allow providers to seamlessly support previous and current versions as long as they update their code before 0.7.0 comes out.

### DIFF
--- a/openfeature/provider/__init__.py
+++ b/openfeature/provider/__init__.py
@@ -1,13 +1,17 @@
+from __future__ import annotations
+
 import typing
+from abc import abstractmethod
 from enum import Enum
 
 from openfeature.evaluation_context import EvaluationContext
+from openfeature.event import ProviderEvent, ProviderEventDetails
 from openfeature.flag_evaluation import FlagResolutionDetails
 from openfeature.hook import Hook
 
 from .metadata import Metadata
 
-__all__ = ["ProviderStatus", "FeatureProvider", "Metadata"]
+__all__ = ["AbstractProvider", "ProviderStatus", "FeatureProvider", "Metadata"]
 
 
 class ProviderStatus(Enum):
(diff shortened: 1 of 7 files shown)


openfeature/provider/__init__.py
```diff
@@ -17,6 +21,20 @@ class ProviderStatus(Enum):
     STALE = "STALE"
     FATAL = "FATAL"
 
+    __events__: typing.ClassVar[typing.Dict[ProviderEvent, str]] = {
+        ProviderEvent.PROVIDER_READY: READY,
+        ProviderEvent.PROVIDER_ERROR: ERROR,
+        ProviderEvent.PROVIDER_ERROR: FATAL,
```
gruebel, line 27: this doesn't work, either we emit `ERROR` or `FATAL` for `PROVIDER_ERROR`

openfeature/provider/__init__.py
```diff
@@ -17,6 +21,20 @@ class ProviderStatus(Enum):
...
+        ProviderEvent.PROVIDER_STALE: STALE,
+    }
+
+    @classmethod
+    def from_provider_event(
+        cls, event: ProviderEvent
+    ) -> typing.Optional[ProviderStatus]:
+        value = ProviderStatus.__events__.get(event)
+        return ProviderStatus[value] if value else None
```
gruebel, line 36:
```suggestion
        return ProviderStatus.__events__.get(event)
```
I don't think you need to invoke `ProviderStatus` again and `.get()` anyway returns `None` if it can't find what you are looking for 🙂

gruebel (commented):
overall it looks good

openfeature/provider/__init__.py
federicobond: good catch, did the change mechanically and forgot it wasn't a bijection

openfeature/provider/__init__.py
federicobond: This was done because the dict values are strings when the dict is defined. In any case, I decided to inline the code to avoid exposing an internal detail

gruebel (approved):

PR #315, opened 2024-04-10
feat: implement transaction context

federicobond:
This is a draft implementation of transaction context propagation via [contextvars](https://docs.python.org/3/library/contextvars.html#module-contextvars).

### DIFF
--- a/openfeature/api.py
+++ b/openfeature/api.py
@@ -12,6 +12,10 @@ from openfeature.hook import Hook
 from openfeature.provider import FeatureProvider
 from openfeature.provider._registry import provider_registry
 from openfeature.provider.metadata import Metadata
+from openfeature.transaction_context import (
+    NoopTransactionContextPropagator,
+    TransactionContextPropagator,
+)
 
 __all__ = [
     "get_client",
@@ -26,11 +30,18 @@ __all__ = [
     "shutdown",
     "add_handler",
     "remove_handler",
+    "set_transaction_context_propagator",
+    "set_transaction_context",
+    "get_transaction_context",
 ]
 
 _evaluation_context = EvaluationContext()
 
 _hooks: typing.List[Hook] = []
+
+_transaction_context_propagator: TransactionContextPropagator = (
+    NoopTransactionContextPropagator()
+)
 
 
 def get_client(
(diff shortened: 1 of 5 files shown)


openfeature/api.py
```diff
@@ -26,12 +30,19 @@
...
+    "set_transaction_context",
+    "get_transaction_context",
 ]
 
 _evaluation_context = EvaluationContext()
 
 _hooks: typing.List[Hook] = []
 
+_transaction_context_propagator: TransactionContextPropagator = (
+    NoopTransactionContextPropagator()
```
federicobond, line 43: Should the SDK initialize a functional transaction context propagator by default?

openfeature/api.py
```diff
@@ -94,3 +105,17 @@ def add_handler(event: ProviderEvent, handler: EventHandler) -> None:
...
+) -> None:
+    _transaction_context_propagator = propagator
+
+
+def set_transaction_context(context: EvaluationContext) -> None:
+    _transaction_context_propagator.set_transaction_context(context)
+
+
+def get_transaction_context() -> EvaluationContext:
+    return _transaction_context_propagator.get_transaction_context()
```
federicobond, line 121: Is this the right API or do we want something based on decorators/context managers?

openfeature/api.py
beeme1mr: We do something similar to this in JS. What would the alternative be?

openfeature/api.py
beeme1mr: Could you elaborate on this? The main use case for accessing transaction context would be within the SDK when it merges context prior to flag evaluation.

openfeature/api.py
federicobond: Having a default implementation that is not a no-op. But this could introduce a performance penalty if you are not using it. Probably better to default to no-op.

openfeature/api.py
federicobond: Yes, the comment should have been in the `set_transaction_context`. I was wondering if we should do something like this instead:

```python
with start_transaction_context(context):
    continue_transaction()  # context will be cleaned up after the block completes
```

Similar to the way it works in JS by taking a callback.

gruebel:
implemented via #389

PR #324, opened 2024-04-30
refactor: bind providers explicitly to a registry with attach/detach

federicobond:
This opens the door to handling events for child providers inside composite providers.

### DIFF
--- a/openfeature/provider/__init__.py
+++ b/openfeature/provider/__init__.py
@@ -23,6 +23,15 @@ class ProviderStatus(Enum):
 
 
 class FeatureProvider(typing.Protocol):  # pragma: no cover
+    def attach(
+        self,
+        on_emit: typing.Callable[
+            [FeatureProvider, ProviderEvent, ProviderEventDetails], None
+        ],
+    ) -> None: ...
+
+    def detach(self) -> None: ...
+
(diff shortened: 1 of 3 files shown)


openfeature/provider/__init__.py
```diff
@@ -68,6 +77,18 @@ def resolve_object_details(
 
 
 class AbstractProvider(FeatureProvider):
+    def attach(
+        self,
+        on_emit: typing.Callable[
+            [FeatureProvider, ProviderEvent, ProviderEventDetails], None
+        ],
+    ) -> None:
+        self._on_emit = on_emit
```
gruebel, line 86: should this work for multiple providers or just a single one?

federicobond: The design takes a lot from the Java SDK implementation.

gruebel: What I meant is, this implementation supports the attachment of one provider, if you run it again, then it will override the previously attached one.

If this is intended, then all is good, if not then we need an internal map to be able to attach multiple.

federicobond: Yes, it supports the attachment of each provider to a single event dispatcher, that is intended. But a single registry can attach to multiple providers.

gruebel (approved):
🍻

beeme1mr (approved):

PR #325, opened 2024-05-01
chore: add Python 3.13 build

gruebel:
## This PR

- a bit early, but next week the first beta release will come out and that marks no further feature development
- also increased the default version to Python 3.12 to align it with the version used during the release


gruebel:
strange error, seems to be `htach` related, because I was able to create a 3.13 venv locally and run our tests there  🤔  will revisit it again with the first RC version in July.

Issue #96, opened 2023-04-18
review thread safety of client/API/Evaluation context


toddbaert:

toddbaert:
@matthewelwell @hlipsig @tcarrio do we need to worry about locking/concurrency here? I know there's a `threading` module in Python. Are any locks necessary, especially around global state? From our meeting if sounds like this might not be necessary, but I have limited Python experience and can't say...

beeme1mr:
Hey @federicobond perhaps you could answer this question. Do we need to be concerned about threading issues?

federicobond:
I'd have to take a closer look at it. The GIL has us covered in some cases for simple assignments but anything more complex than that should probably be guarded by a lock.

`add_hooks` is probably not thread-safe due to read-update-write race conditions when concatenating lists.

`set_provider` used to be thread safe but likely isn't now with the new shutdown and initialize logic.

toddbaert:
Did an audit of this as part of looking into 1.0 readiness (#522). Updated the description with findings.

tl;dr: the only locks in the SDK are in `_event_support.py`. Everything else (registry, hooks, eval context, transaction context, client state, provider `_on_emit`) is unprotected. The `ProviderRegistry` is the worst offender since it has multi-step mutations with side effects (init/shutdown), but there are races in most shared mutable state.

See the description above for the full list.

leakonvalinka:
I'll start working on this now!

PR #331, opened 2024-05-04
fix: make global hooks thread safe

gruebel:
## This PR
- I tried to create a similar test case like in #329, but couldn't make it work 🙁 
- also removed unneeded `global` keyword, only one is left, which can't be removed

### Related Issues

Relates #96


beeme1mr (approved):

openfeature/api.py
```diff
@@ -70,18 +71,18 @@ def set_evaluation_context(evaluation_context: EvaluationContext) -> None:
 
 
 def add_hooks(hooks: typing.List[Hook]) -> None:
-    global _hooks
-    _hooks = _hooks + hooks
```
federicobond, line 74: Variable assignment should be atomic thanks to the GIL. Not sure I understand the potential race conditions here.

openfeature/api.py
```diff
@@ -30,6 +31,7 @@
 
 _evaluation_context = EvaluationContext()
 
+_hooks_lock = threading.RLock()
```
federicobond, line 34: If we know we are not going to be reentrant, it's faster to use a `threading.Lock`

openfeature/api.py
gruebel: that's correct, but you have unpredictable behaviour when multithreaded, because the `_hooks + hooks` creates a complete new list, which has nothing to do with the old one. While locking this part all hooks are added and you can be sure they will all be triggered with any of the hook functions.

openfeature/api.py
gruebel: yeah, let's use a normal `Lock` I don't think it is going to be reentrant. If it changes, we can easily change it to `RLock` 😄

openfeature/api.py
federicobond: But isn't that a good thing? Having each version of the list effectively immutable simplifies sharing data safely across threads. With a mutable list I can iterate over the hooks inside the client (after the `_hooks_lock` is released from `get_hooks`) while the list is being modified inside `add_hooks` or `clear_hooks`.

gruebel: I get what you mean, I just feel like my approach will result in the same behaviour as the `providers` code. Both approaches have their benefits.

beeme1mr:
Hey @federicobond, could you please re-review this when you have a moment? I would like to cut a 1.0 release candidate soon.

openfeature/api.py
matthewelwell: Yeah, I certainly see @federicobond's point here. In code, I guess what he's saying is that some thread is doing something like: 

```python
for hook in get_hooks():
    do_something()
```

while some other thread is doing something like: 

```python
add_hook(MyHook())
```

> Variable assignment should be atomic thanks to the GIL. Not sure I understand the potential race conditions here.

@federicobond I think this stemmed from your comment [here](https://github.com/open-feature/python-sdk/issues/96#issuecomment-1769569433), but I actually think your comment on _this_ PR is correct, that the GIL would prevent these race conditions from occurring. In which case, I think the current implementation is safer than the one implemented in this PR?

gruebel: both approaches should be similar thread safe. Using the lock gives better predictability about the actual behaviour as mentioned. If you add a hook in one thread and get all hooks in another thread, then you will either get all + the one you added or not, depends on how fast Python was able to create the new list and replaced it. Also looking at the current development in Python 3.13 you get the possibility to turn off the GIL 😱 

I talked with @federicobond a while back and we keep the current implementation I'll just remove the unneeded `global` calls in 2 cases.

matthewelwell (approved):

openfeature/api.py
matthewelwell: Thanks @gruebel , I have approved the PR.

PR #340, opened 2024-06-06
ci: update release please to new GHA

### LINKED ISSUE #332: Expose SDK version as a module constant for use by providers/hooks
It would be useful to expose a module constant with the current SDK version so that providers can include it in User-Agent headers for example.

gruebel:
## This PR

- the `google-github-actions/release-please-action` GHA was deprecated and this the replacement 🙂 
- also added the `version.py` file, which hopefully will get updated correctly with the SDK version and can be used by 3rd parties

### Related Issues

Fixes #332


openfeature/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.7.0"
```
beeme1mr, line 1: What is this for?

beeme1mr (approved):

openfeature/version.py
gruebel: this gives the possibility to import the SDK version on runtime and add it to the user agent. nothing crucial for the release itself.

beeme1mr: Should we have Release Please manage this?

gruebel: Yeah, I think it should automatically do that, at least that's how I understood their documentation.

beeme1mr: Okay, sounds good. Let's give it a shot.

PR #343, opened 2024-06-16
chore: add Python 3.13

gruebel:
- let's see if it now works better, than a few weeks ago 🧐


gruebel:
next try in 1-2 months again 😅

PR #347, opened 2024-07-01
fix: remove exception logging during evaluation

beeme1mr:
## This PR

- remove exception logging during evaluation

### Impacts

https://github.com/open-telemetry/opentelemetry-demo/issues/1628

### Notes

The OTel demo was seeing overly verbose log messages when they disabled a flag. That's because a `FLAG_NOT_FOUND` error was raised by the provider, and the SDK was logging all exceptions. I've updated the SDK to no longer log OpenFeatureErrors since providers commonly raise these, which can lead to log spam. It's worth noting that this information can still be accessed via a hook or evaluation details.

### DIFF
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -361,19 +361,13 @@ class OpenFeatureClient:
             return flag_evaluation
 
         except OpenFeatureError as err:
-            logger.exception(
-                "Error %s while evaluating flag with key: '%s'",
-                err.error_code,
-                flag_key,
-            )
-
             error_hooks(flag_type, hook_context, err, reversed_merged_hooks, hook_hints)
 
             return FlagEvaluationDetails(
                 flag_key=flag_key,
                 value=default_value,
                 reason=Reason.ERROR,
-                error_code=err.error_code,
+                error_code=err.error_code or ErrorCode.GENERAL,
                 error_message=err.error_message,
             )
         # Catch any type of exception here since the user can provide any exception
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -367,7 +367,7 @@ class OpenFeatureClient:
                 flag_key=flag_key,
                 value=default_value,
                 reason=Reason.ERROR,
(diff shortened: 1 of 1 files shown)


openfeature/client.py
```diff
@@ -361,19 +361,13 @@ def evaluate_flag_details(  # noqa: PLR0915
...
-            )
-
             error_hooks(flag_type, hook_context, err, reversed_merged_hooks, hook_hints)
 
             return FlagEvaluationDetails(
                 flag_key=flag_key,
                 value=default_value,
                 reason=Reason.ERROR,
-                error_code=err.error_code,
+                error_code=err.error_code or ErrorCode.GENERAL,
```
gruebel, line 370: This shouldn't be needed, because every OF error has an error code.

dyladan:
Is this affecting only the Python SDK or is it something that should be specified more generally?

beeme1mr:
> Is this affecting only the Python SDK or is it something that should be specified more generally?

Logging isn't covered in the spec. We could consider general guidance but I would be reluctant to go any deeper than that.

openfeature/client.py
beeme1mr: Okay, I can remove it. We just did something similar in JS and the typings implied that this was optional.

toddbaert (approved):

toddbaert:
> Is this affecting only the Python SDK or is it something that should be specified more generally?

After approving, I'm actually considering if we should specify something here. I'm wondering if this is the wrong solution to the problem.

I think we _should_ be logging errors if a flag is not found, and that this change will make debugging harder. I think this is a situation where we'd want log-spam.

toddbaert:
The Java SDK throws here: https://github.com/open-feature/java-sdk/blob/fc40209edcffc063a474aec7bfe9a880e0966750/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L135C1-L136C1

PR #385, opened 2024-11-13
feat: simplify async functionality to provider

leohoare:
## This PR
Adds the ability for open feature providers to use async methods
Note: This version simplifies the implementation and doesn't change any of the hook implementation.

### Related Issues
https://github.com/open-feature/python-sdk/issues/284
https://github.com/open-feature/python-sdk/pull/383

### Notes
Just confirming the approach / POC for how this would work.


beeme1mr:
Hey @leohoare, thanks for the pr. I'm at KubeCon this week so reviews may be a bit delayed but I'll take a look as soon as possible.

leohoare:
No rush @beeme1mr, enjoy kubecon.

openfeature/api.py
```diff
@@ -39,6 +39,12 @@ def get_client(
     return OpenFeatureClient(domain=domain, version=version)
 
 
+def get_client_async(
```
gruebel, line 42:
```suggestion
def get_async_client(
```
I would prefer this name, because the current one somehow implies I'm getting the client asynchronously, at least for me 😅

openfeature/provider/in_memory_provider.py
```diff
@@ -117,3 +117,56 @@ def _resolve(
...
+    _flags: FlagStorage
+
+    def __init__(self, flags: FlagStorage) -> None:
+        self._flags = flags.copy()
+
+    def get_metadata(self) -> Metadata:
+        return InMemoryMetadata()
+
+    def get_provider_hooks(self) -> typing.List[Hook]:
+        return []
```
gruebel, line 132: maybe I'm wrong, but I think these two should also be `async` 🤔

gruebel (commented):
overall nice work 🍻

openfeature/api.py
leohoare: Agreed, a bit of an oversight.

openfeature/provider/in_memory_provider.py
leohoare: Do you mean more generally or specifically to the InMemoryProvider?

For this PR, I was limiting the scope of adding async functionality to the resolving calls. 
This ensures we have backwards compatibility with current hooks / metadata calls and providers would just need to implement the async resolve calls.

I do agree support async hooks and metadata retrieval should be looked at in the future though.

gruebel: Yeah, meant it more generally, because currently those would block the call, which is not really relevant for the in-memory variant, but would be for remote calls.

But totally fine to change this in the future.

leohoare:
The current approach wasn't very compatible with mypy, I've updated settings to ignore overrides.
i.e. 
```
    def get_boolean_value(
        ...
    ) -> bool:
```
Can be overwritten to 
```
    async def get_boolean_value(
        ...
    ) -> bool:
```
Also given we're not forking the typing protocol `FeatureProvider` the following line isn't recognised as awaitable.
```
resolution = await get_details_callable(*args)
```
So I've ignored the error for that line. `# type: ignore[misc]`

I'm not a heavy mypy user, let me know if there's a better or cleaner approach you'd prefer here.

beeme1mr (commented):
What's the expected behavior if a user registers an async provider but uses the synchronous client and vice versa? Could you please update the root readme examples for how to use this? If providers can't be used interchangeably, it may fragment the Python-related ecosystem in a confusing way.

openfeature/client.py
```diff
@@ -464,3 +465,353 @@ def _typecheck_flag_value(value: typing.Any, flag_type: FlagType) -> None:
...
+            FlagType.FLOAT: provider.resolve_float_details,
+            FlagType.OBJECT: provider.resolve_object_details,
+            FlagType.STRING: provider.resolve_string_details,
+        }
+
+        get_details_callable = get_details_callables.get(flag_type)
+        if not get_details_callable:
+            raise GeneralError(error_message="Unknown flag type")
+
+        resolution = await get_details_callable(*args)  # type: ignore[misc]
```
ChihweiLHBird, line 803: Might be helpful to check [whether `get_details_callable` is awaitable](https://docs.python.org/3/library/inspect.html#inspect.isawaitable) here in order to make sync provider compatible with async clients?

It might look like this, but I didn't test it
```suggestion
        resolution = get_details_callable(*args)  # type: ignore[misc]
        if isawaitable(resolution):
            resolution = await resolution
```

ChihweiLHBird:
@beeme1mr I think it's generally hard to perfectly call an async functions (from a provider) in a sync function (from a client) due to possibility of customized or already-started asyncio event loop, but calling sync function from async functions are much easier.

I would still recommend implementing the async support but we do need some error checking and documentation about incompatibility between sync clients and async providers.

beeme1mr:
Hey @ChihweiLHBird, I'm definitely in favor of adding async support. I'm just familiar enough with Python to provide a solid recommendation here. My main concern is putting too much burden on the end user to understand what provider and hook are compatible with a particular client.

@federicobond @gruebel @aepfli do you have any recommendations?

beeme1mr:
Hey @leohoare, sorry for the delay and Happy New Year! This fell off my radar. Feel free to @ mention me if it happens again.

ChihweiLHBird:
@beeme1mr Thanks! And I just realized we may still make the async provider be compatible with sync client. This pattern below may be helpful to determine if the current execution is in an `asyncio` event loop. So I think we should be able to make it 2 ways compatible in most cases.

(Need more opinions on this pattern since I am not very familiar with it, I'm not sure if it's 100% reliable/secure)

```python
import asyncio

try:
    asyncio.get_running_loop()
except RuntimeError:
    asyncio.run(some_io_task())
```

ChihweiLHBird:
@beeme1mr Thanks! And I just realized we may still make the async provider be compatible with sync client. This pattern below may be helpful to determine if the current execution is in an `asyncio` event loop. So I think we should be able to make it 2 ways compatible in most cases.

(Need more opinions on this pattern since I am not very familiar with it, I'm not sure if it's 100% reliable/secure)

```python
import asyncio

try:
    asyncio.ensure_future(asyncio.get_running_loop().create_task(some_async_io_task()))
except RuntimeError:
    asyncio.run(some_async_io_task())
```

But generally, a similar effort will get extremely complicated, and it might be better and easier to just require asyncs client for asyncs provider or sync clients for sync providers.

ChihweiLHBird:
@leohoare Most project I have seen implemented async stuff as you described, and I am also personally leaning to the style other projects used for our SDK here.

aepfli:
Hey, sorry that i chime in here too. For me this seems like something, which we might should discuss even on a "global" open feature level. There are other languages, which also support async functionalities (like java with Futures) and maybe we should write a proper specification, on how we want to handle this. I think it could be problematic, if we do have mutliple different ways within our eco system to handle async. Hence it is worth to shortly discuss this on a bigger level. wdty? (@open-feature/technical-steering-committee fyi)

beeme1mr:
I think it makes sense to include the async methods on the existing client. It should be possible for the async calls to be accessed via the async method (even if they're not truly async), meaning exiting providers would "just work." How would we want to support async-only providers? Keep in mind that the evaluation can't throw an error that bubbles up to the user.

openfeature/client.py
lukas-reining: I actually think this has to be done. If not I think we would get type errors right?

lukas-reining (commented):
I am not a huge python expert, but I have a concern regarding the hard split between the sync and async world here.
If I am not seeing it wrong, a provider would either implement the sync or async world or both in this design. Is this correct?

If this is true: Do you see an option to have a sync provider and client while having an async client that wraps the sync client similar to:

```python
class OpenFeatureClient:
    def _evaluate_flag_details(
        self,
        flag_type: FlagType,
        flag_key: str,
        default_value: Any,
        evaluation_context: Optional[EvaluationContext] = None,
        flag_evaluation_options: Optional[FlagEvaluationOptions] = None,
    ) -> FlagEvaluationDetails[Any]:
        # Dummy implementation
        return FlagEvaluationDetails()

    def get_boolean_details(self,
        flag_key: str,
        default_value: bool,
        evaluation_context: typing.Optional[EvaluationContext] = None,
        flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None) -> FlagEvaluationDetails[bool]:
        return self._evaluate_flag_details(FlagType.BOOLEAN, flag_key, default_value, evaluation_context, flag_evaluation_options)

    async def _async_wrap(self, func: Callable, *args, **kwargs) -> typing.Any:
        loop = asyncio.get_event_loop()
        return await loop.run_in_executor(None, partial(func, *args, **kwargs))

    async def get_boolean_details_async(self, 
        flag_key: str, 
        default_value: bool, 
        evaluation_context: typing.Optional[EvaluationContext] = None, 
        flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None) -> FlagEvaluationDetails[bool]:
        return await self._async_wrap(
            self.get_boolean_details, 
            flag_key, 
            default_value, 
            evaluation_context, 
            flag_evaluation_options
        )
```

The types and params might be a bit off, I just wanted to show the idea. 
@open-feature/sdk-python-maintainers maybe you also have an opinion to this.

I would like to avoid "double" amount of work for implementers of providers if this is possible. 

cc @gruebel @beeme1mr

lukas-reining:
> Hey, sorry that i chime in here too. For me this seems like something, which we might should discuss even on a "global" open feature level. There are other languages, which also support async functionalities (like java with Futures) and maybe we should write a proper specification, on how we want to handle this. I think it could be problematic, if we do have mutliple different ways within our eco system to handle async. Hence it is worth to shortly discuss this on a bigger level. wdty?

This conforms to what I also wrote in my comment I just wrote @aepfli.
E.g. async/await is really different between JS and Python in my view. But still I think we might be able to come up with some general things like "avoiding double implementation" if possible.

lukas-reining:
> I think it makes sense to include the async methods on the existing client. It should be possible for the async calls to be accessed via the async method (even if they're not truly async), meaning exiting providers would "just work." How would we want to support async-only providers? Keep in mind that the evaluation can't throw an error that bubbles up to the user.

I am totally with you with this @beeme1mr. 

The only thing I think I would question is `I think it makes sense to include the async methods on the existing client.`, as in the past I have seen several libs that implement explicit async and sync classes while sharing the implementation of the core logic. But tbh. I would accept whatever a more Python familiar person than me says is more common or "idiomatic".

leohoare:
Agree with all the above conversation, to summarise:
* We should use single client and add _async methods
* Async methods should share common code with sync
* If a provider only has sync calls, async methods should "work" and use these methods

I'll have a think how to do this in python but I'm on the same page with the current code has way too much duplication. 
It sounds like it somewhat needs to "check" if an async method exists, if not fallback to sync. At that point it would need to optionally await the call. I'll see if we could then "bubble" this up to separate async methods.

I'm not sure how this could be extended to support async hooks though.

toddbaert:
> I'm not sure how this could be extended to support async hooks though.

hmmm... we might be forced to make all hooks async in this case, which is the case for server-side JS: https://github.com/open-feature/js-sdk/blob/main/packages/server/src/hooks/hook.ts

That way they will work in either case... does that make sense? We are still pre-1.0 so now is the time if we need to make this sort of change.

leohoare:
### Fallback Sync Functionality
To achieve falling back on to the sync methods and not forcing async implementation. Adding default method to the Abstract provider is one approach. The downside being it wouldn't support only async implementations. It also wouldn't be an `@abstractmethod`, so we wouldn't be forcing the implementation.
```python
class AbstractProvider(FeatureProvider):
    ...
    async def resolve_boolean_details_async(
        self,
        flag_key: str,
        default_value: bool,
        evaluation_context: typing.Optional[EvaluationContext] = None,
    ) -> FlagResolutionDetails[bool]:
        return self.resolve_boolean_details(flag_key, default_value, evaluation_context)
```
### Async/Sync Functionality
It's hard to achieve what we're going for without some big drawbacks.
#### Async wrapping
Idea from @lukas-reining 
```python
    async def _async_wrap(self, func: Callable, *args, **kwargs) -> typing.Any:
        loop = asyncio.get_event_loop()
        return await loop.run_in_executor(None, partial(func, *args, **kwargs))

    async def get_boolean_details_async(self, 
        flag_key: str, 
        default_value: bool, 
        evaluation_context: typing.Optional[EvaluationContext] = None, 
        flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None) -> FlagEvaluationDetails[bool]:
        return await self._async_wrap(
            self.get_boolean_details, 
            flag_key, 
            default_value, 
            evaluation_context, 
            flag_evaluation_options
        )
```
My main concerns:
* It will offload the underlying call to a thread, which impact scalability 
* my gut feeling is we could potentially introduce deadlocks if the underlying call is async
* we're not really leveraging the advantages of async programming

#### Commonly merging async/sync
A pattern used where entrypoints are async/sync (client), calling sync common code (client) and then forking logic for async/async (provider).
```python
    def get_object_details_async(
        self,
        flag_key: str,
        default_value: typing.Union[dict, list],
        evaluation_context: typing.Optional[EvaluationContext] = None,
        flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None,
    ) -> FlagEvaluationDetails[typing.Union[dict, list]]:
        return self.evaluate_flag_details(
            FlagType.OBJECT,
            flag_key,
            default_value,
            evaluation_context,
            flag_evaluation_options,
            True, # flag for if async/sync
        )
...
  def _create_provider_evaluation(
        if asyncio.iscoroutinefunction(get_details_callable):
            resolution = asyncio.run(get_details_callable(*args))
        else:
            resolution = get_details_callable(*args)
```
This won't work because:
* async calls will throw a runtime error when the event loop is already running (e.g. pytest, APIs etc)
* a bit hacky

#### Converting downstream functions to all async
```python
    def get_integer_details(
        self,
        flag_key: str,
        default_value: int,
        evaluation_context: typing.Optional[EvaluationContext] = None,
        flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None,
    ) -> FlagEvaluationDetails[int]:
        return asyncio.run(
            self.evaluate_flag_details(
                FlagType.INTEGER,
                flag_key,
                default_value,
                evaluation_context,
                flag_evaluation_options,
            )
        )

    async def get_integer_details_async(
        self,
        flag_key: str,
        default_value: int,
        evaluation_context: typing.Optional[EvaluationContext] = None,
        flag_evaluation_options: typing.Optional[FlagEvaluationOptions] = None,
    ) -> FlagEvaluationDetails[int]:
        return await self.evaluate_flag_details(
            FlagType.INTEGER,
            flag_key,
            default_value,
            evaluation_context,
            flag_evaluation_options,
        )

    async def evaluate_flag_details(
          ...
```
Although this works as soon as we start running an event loop, we'll error out (APIs, async pytest).
This would be a breaking change for anyone wrapping sync calls in async code.

#### Abstracting common code
Which leaves us the final option, abstracting common code and forking the calls, similar to what I've done but without duplicating the code base.
This seems like the most promising option at the moment with no obvious downsides apart from code maintainability and upfront work.
Theoretically, there shouldn't be an issue extending to async hooks.

leohoare:
Closing in favour of https://github.com/open-feature/python-sdk/pull/413.

Issue #292, opened 2024-03-13
[FEATURE] Implement transaction context


toddbaert:
### Requirements

# We should implement transaction context propagation

## What is transaction context?

[From the spec:](https://openfeature.dev/specification/sections/evaluation-context#33-context-propagation)
Transaction context is a container for transaction-specific evaluation context (e.g. user id, user agent, IP).
Transaction context can be set where specific data is available (e.g. an auth service or request handler) and by using the transaction context propagator it will automatically be applied to all flag evaluations within a transaction (e.g. a request or thread).

Transaction context has been added in the following PR: https://github.com/open-feature/spec/pull/227

### Implementation
The JS SDK implementation is implemented as a POC that is currently being hardened.
https://github.com/open-feature/js-sdk/blob/main/packages/server/src/transaction-context/transaction-context.ts
We'd need to define some interfaces to support the construction various context propagation mechanisms, for instance, one based on the [flask context](https://flask.palletsprojects.com/en/2.3.x/reqcontext/).

### Usage
An example usage for the JS SDK implementation of the feature is the usage for propagating HTTP request information in the NestJS SDK.
https://github.com/open-feature/js-sdk/blob/main/packages/nest/src/evaluation-context-interceptor.ts#L31

federicobond:
We could prototype a propagator based on [contextvars](https://docs.python.org/3/library/contextvars.html#module-contextvars) which are sort of threadlocals that also work well in async contexts. This would be framework-agnostic and should be able to work well with both Django and Flask.

federicobond:
Must the API be designed to lexically scope the transaction context, as in a context manager, so that it cleans up after itself? Or is it permissible to set the transaction evaluation context and return, leaving the cleanup for a separate method?

What happens if it's called nested? Does the nested context override the previous and then restores it on exit?

lukas-reining:
I started an implementation similar to the ones in Node and Java, will try to finish tomorrow :)

PR #389, opened 2024-11-16
feat: implement transaction context

lukas-reining:
Implements transaction context as described in #292.
This is done the same as in Node, the callback we have for Node is not needed with context vars, like it is in Java.

@federicobond just after I did the changes I saw you PR from May:
Sorry, I could have just continued test on this one. 
We basically did the same but I added some tests also for the context merging.

### DIFF
--- a/openfeature/transaction_context/__init__.py
+++ b/openfeature/transaction_context/__init__.py
@@ -1,0 +1,15 @@
+from openfeature.transaction_context.context_var_transaction_context_propagator import (
+    ContextVarsTransactionContextPropagator,
+)
+from openfeature.transaction_context.no_op_transaction_context_propagator import (
+    NoOpTransactionContextPropagator,
+)
+from openfeature.transaction_context.transaction_context_propagator import (
+    TransactionContextPropagator,
+)
+
+__all__ = [
+    "TransactionContextPropagator",
+    "NoOpTransactionContextPropagator",
+    "ContextVarsTransactionContextPropagator",
+]
--- a/openfeature/transaction_context/__init__.py
+++ b/openfeature/transaction_context/__init__.py
@@ -1,15 +1,11 @@
 from openfeature.transaction_context.context_var_transaction_context_propagator import (
     ContextVarsTransactionContextPropagator,
 )
-from openfeature.transaction_context.no_op_transaction_context_propagator import (
-    NoOpTransactionContextPropagator,
-)
 from openfeature.transaction_context.transaction_context_propagator import (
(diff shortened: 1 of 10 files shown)


beeme1mr (approved):

openfeature/transaction_context/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from openfeature.transaction_context.no_op_transaction_context_propagator import (
+    NoOpTransactionContextPropagator,
+)
+from openfeature.transaction_context.transaction_context_propagator import (
+    TransactionContextPropagator,
+)
+
+__all__ = [
+    "TransactionContextPropagator",
+    "NoOpTransactionContextPropagator",
```
gruebel, line 13: to keep it aligned with the other `NoOp` classes I would prefer not to expose it here and when needed just import it via the `no_op_transaction_...` file directly.

openfeature/transaction_context/no_op_transaction_context_propagator.py
```diff
@@ -0,0 +1,13 @@
...
+)
+
+
+class NoOpTransactionContextPropagator(TransactionContextPropagator):
+    def get_transaction_context(self) -> EvaluationContext:
+        return EvaluationContext()
+
+    def set_transaction_context(
+        self, transaction_context: EvaluationContext
+    ) -> None: ...
```
gruebel, line 13: also here to be consistent with the code, just use `pass`
```suggestion
    ) -> None:
    pass
```

`...` in `Protocol` is correct 🙂

openfeature/transaction_context/transaction_context_propagator.py
```diff
@@ -0,0 +1,14 @@
+import typing
+from typing import TypeVar
+
+from openfeature.evaluation_context import EvaluationContext
+
+T = TypeVar("T", bound="TransactionContextPropagator")
```
gruebel, line 6: Maybe you planed to somehow use it, but I don't see anything. please elaborate or remove it.

gruebel (commented):
nice work, just added a few minor comments

openfeature/transaction_context/transaction_context_propagator.py
lukas-reining: Yes it was from before, thanks!

gruebel (approved):

gruebel:
I'll give @federicobond till end of week to leave some comments, otherwise will merge this and close the previous draft.

PR #423, opened 2025-01-28
feat: Add evaluation details to finally hook stage #403

### LINKED ISSUE #403: Add `evaluation details` to `finally` hook stage
Add `evaluation details` to `finally` hook stage.

### Notes

This breaks the signature of the `finally` stages based on [this spec enhancement](https://github.com/open-feature/spec/pull/280). It is **not** considered a breaking change to the SDK because hooks are marked as experimental in the spec, and the change has no impact on known hooks.

```psuedo
- finally(hookContext, hookHints)
+finally(hookContext, evaluationDetails, hookHints)
```

## Acceptance criteria: 

- finally hook stage modified
- associated tests

## Follow ups:

- update contribs

chrfwow:
## This PR
Adds the evaluation details to the finally hook

### Related Issues

Fixes #403 

### How to test
Please make sure the e2e tests work, especially the part with `git submodule add --force https://github.com/open-feature/spec.git spec` in `pyproject.toml`.

### DIFF
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -393,7 +393,13 @@ class OpenFeatureClient:
             return flag_evaluation
 
         finally:
-            after_all_hooks(flag_type, hook_context, flag_evaluation, reversed_merged_hooks, hook_hints)
+            after_all_hooks(
+                flag_type,
+                hook_context,
+                flag_evaluation,
+                reversed_merged_hooks,
+                hook_hints
+            )
 
     def _create_provider_evaluation(
         self,
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -325,7 +325,6 @@ class OpenFeatureClient:
                 error_code=ErrorCode.PROVIDER_FATAL,
             )
 
-        flag_evaluation = None
         try:
             # https://github.com/open-feature/spec/blob/main/specification/sections/03-evaluation-context.md
             # Any resulting evaluation context from a before hook will overwrite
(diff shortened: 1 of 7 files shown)


chrfwow:
The code coverage report fails because one line is not covered by tests. However, this line should not be reachable and therefore cannot be tested easily. I only added it to make the linter happy.

openfeature/client.py
```diff
@@ -325,6 +325,7 @@ def evaluate_flag_details(  # noqa: PLR0915
                 error_code=ErrorCode.PROVIDER_FATAL,
             )
 
+        flag_evaluation = None
```
gruebel, line 328: if I understand the spec correctly then the finally hook should also return evaluation details on success, right? if yes, then you can remove this and also the `if flag_evaluation is None: ...` block under `finally`

chrfwow: I don't think I can. The variable `flag_evaluation` needs to be assigned some value in any case, and this in theory might not happen when the finally block is executed. Therefore I need to assign `None` initially, and then do the check if the value is still `None`, even though I know that it never will be.

gruebel: I tested it locally and it works without any problems, both tests and linting passes. The reason why it works, the bad cases you covered in this PR and the good case already assigns the evaluation detail response to the same variable `flag_evaluation`.

gruebel (approved):
Thanks 🍻

openfeature/client.py
chrfwow: The build completes fine, but IntelliJ gives a (false positive) warning

gruebel: Yeah, not the first time an IDE can't fully comprehend code 😞

gruebel:
if there are no other objections, then I will merge end of week.

beeme1mr (approved):

toddbaert (approved):
My Python is not great, but this looks spec compliant to me.

lukas-reining (approved):
Looks good to me!

PR #433, opened 2025-02-05
chore: use centralized renovate config, downgrade release please

beeme1mr:
Extends the [centralized Renovate config](https://github.com/open-feature/community-tooling/blob/main/renovate.json) and downgrades Release Please to v3.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: googleapis/release-please-action@v4
+      - uses: googleapis/release-please-action@v3
         id: release
         with:
           command: manifest
--- a/renovate.json
+++ b/renovate.json
@@ -1,30 +1,12 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended"
-  ],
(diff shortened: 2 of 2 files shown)


toddbaert (approved):

.github/workflows/release.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: googleapis/release-please-action@v4
+      - uses: googleapis/release-please-action@v3
```
gruebel, line 24: Just curious why we need to downgrade? Are there some bugs or it doesn't work well with our setup?

gruebel (approved):

.github/workflows/release.yml
beeme1mr: Yeah, unfortunately. There are two issues I'm aware of.

- https://github.com/googleapis/release-please-action/issues/904
- https://github.com/googleapis/release-please-action/issues/912

We've implemented the workaround for the second issue but the first is more difficult to get around. Our release PR will always fail the DCO check. See [here](https://github.com/open-feature/python-sdk/pull/431).

gruebel: Gotcha, saw that happening on the past. Good to know what caused it, thanks!

PR #441, opened 2025-02-07
chore!: drop Python 3.8 support

gruebel:
## This PR

- I think it is time to say goodbye to Python 3.8 support, `mypy` dropped 3.8 support in its latest version and the flagd test suite will benefit from it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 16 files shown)


gruebel:
@beeme1mr should we release it as 0.8 version or is 0.7.6 ok too?

beeme1mr:
I'm fine with that and recommending keeping it a breaking change so we can it's very obvious in the release notes.

I'm just wondering if we should release the async changes before merging this.

beeme1mr:
Python 3.8 EOL was 2024-10-07.

https://devguide.python.org/versions/

beeme1mr (approved):

gruebel:
> I'm just wondering if we should release the async changes before merging this.

@beeme1mr yeah I thought about the same thing. I'm somehow more inclined to put the async change with this together into the first 0.8.0 version. It is technically not a breaking change, but still quite a big new feature.

beeme1mr:
Sure, either way works for me. We should give folks a few days to let us know if dropping 3.8 is fine. Let's aim for merging this early next week unless we run into any blockers.

lukas-reining (approved):

leohoare (approved):
out with the old

chrfwow (approved):

PR #444, opened 2025-02-11
chore: use existing submodule version for e2e tests

beeme1mr:
## This PR

- modifies the submodule set in the E2E tests to use the existing `.gitmodule` configuration.

### Notes

The E2E tests will now use the submodule sha instead of pulling the latest. As a added benefit, Renovate should automagically open PRs when the spec repo is updated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ cov = [
     "cov-report",
 ]
 e2e = [
-    "git submodule add --force https://github.com/open-feature/spec.git spec",
+    "git submodule update --init --recursive",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -48,7 +48,7 @@ cov = [
     "cov-report",
 ]
 e2e = [
-    "git submodule add --force https://github.com/open-feature/spec.git spec",
+    "git submodule update --init --recursive",
```
beeme1mr, line 51: This isn't necessary when running in a GitHub action but it's probably worth keeping for local development.

PR #451, opened 2025-02-16
feat: add OTel utility function

### LINKED ISSUE #448: Add OpenTelemetry-Compatible Telemetry Utility Function
## Description

Add a telemetry utility function to the SDK that can be used within the [finally stage](https://openfeature.dev/specification/sections/hooks/#requirement-442) of a hook to create OpenTelemetry-compatible telemetry signals. It must comply with the OpenTelemetry specification and meet the technical requirements, including those from [Appendix D](https://openfeature.dev/specification/appendix-d/) of the OpenFeature Specification and the [OpenTelemetry Semantic Conventions for Feature Flags](https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/).

## Requirements

- Must comply with the OpenTelemetry (OTel) specification.
- Must be included in the SDK.
- Must not include any third-party dependencies.
- Must adhere to the observability conformance requirements in Appendix D of the OpenFeature Specification.

## Example implementation

A JavaScript implementation can found [here](https://github.com/open-feature/js-sdk/pull/1120).

gruebel:
## This PR

- it is mostly equivalent to the JS implementation, except I named the field `body` and not `data`, which seems to fit better to OTel

### Related Issues

Fixes #448

### DIFF
--- a/openfeature/exception.py
+++ b/openfeature/exception.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import typing
 from collections.abc import Mapping
-from enum import Enum
+
+from openfeature._backports.strenum import StrEnum
 
 __all__ = [
     "ErrorCode",
@@ -163,7 +164,7 @@ class InvalidContextError(OpenFeatureError):
         super().__init__(ErrorCode.INVALID_CONTEXT, error_message)
 
 
-class ErrorCode(Enum):
+class ErrorCode(StrEnum):
     PROVIDER_NOT_READY = "PROVIDER_NOT_READY"
     PROVIDER_FATAL = "PROVIDER_FATAL"
     FLAG_NOT_FOUND = "FLAG_NOT_FOUND"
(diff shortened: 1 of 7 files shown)


beeme1mr:
It looks like the OTel Python sdk only supports attributes in both span and log events. @dyladan can we include value as an attribute?

dyladan:
I agree `body` is a better term because it is used by the otel spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#log-and-event-record-definition

All language implementations in OTel are experimental so I would refer to spec before referring to any language terms used.

beeme1mr:
> I agree `body` is a better term because it is used by the otel spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#log-and-event-record-definition
> 
> All language implementations in OTel are experimental so I would refer to spec before referring to any language terms used.

I've updated the JS implementation to use `body`.

beeme1mr (approved):
Look great, thanks! Good catch on the data vs body. I was just looking at the JS implementation but body is a better choice.

PR #480, opened 2025-04-06
fix: fix cycle dependency between api and client

### LINKED ISSUE #478: Importing `OpenFeatureClient` leads to an error
If for any reason I import `OpenFeatureClient` directly without having previously imported `api` in the same file I receive the following error message:

```
ImportError: cannot import name 'OpenFeatureClient' from partially initialized module 'openfeature.client' (most likely due to a circular import) (.venv/lib/python3.11/site-packages/openfeature/client.py)
```

A minimal bit of code that replicates this:

```
from openfeature.client import OpenFeatureClient
```

This can be worked around by always importing the API first:

```
from openfeature import api
from openfeature.client import OpenFeatureClient
```

Is this expected?  Am I not supported to be interacting with the client directly?  I'm using it for typing information & dependency injection definitions.

gruebel:
## This PR

- moved the api function related to hooks, evaluation context and transaction context to their dedicated module and re-exposed them, so there is no difference in the import statement, but now you can directly import them without needing to do it via `openfeature.api`
- I didn't find a good way to make sure we don't this again. I know of `importlinter` which could help, but it doesn't work on the base of the namespace package.

### Related Issues

Fixes #478


openfeature/provider/_registry.py
```diff
@@ -66,8 +66,6 @@ def shutdown(self) -> None:
 
     def _get_evaluation_context(self) -> EvaluationContext:
         # imported here to avoid circular imports
```
imnotjames, line 68: I think this comment can go, too - it was only relevant to the import

imnotjames (commented):
I grabbed the code from this PR and as far as I can tell it resolves the issue I was running into

beeme1mr (approved):

openfeature/provider/_registry.py
gruebel: very true 😅

aepfli (approved):

PR #504, opened 2025-05-31
refactor: refine typing.Any type hints

gruebel:
## This PR

- replaced most of the `Any` type usage with proper types


aepfli (approved):
Thank you!

leohoare (commented):
👍🏻

aepfli:
i think we should have marked this as a minor change, it is breaking the api ;( - learning by doing

Issue #502, opened 2025-05-30
[TOOLING] migrate to uv


aepfli:
We recently adopted the openfeature Python SDK within our company and added a wrapper around the flagd provider.
We discovered `uv` for setting up the project and handling all the build-related tools, which felt nice, easy, and really comfortable. So far, everyone loves it from the tooling side, and it also seems like the current way to go within the Python universe, with `ty` on the horizon as a fast type checker.

The migration from `hatch` to `UV` should be pretty straightforward, as `UV' also uses hatchling by default.

I suggest that we migrate to UV within our Python packages to make adaptation and running the project easier. This issue is to talk about it, and discuss it. There are for sure some trade-offs, but so far I highly enjoyed working with `uv` way more than working with `hatch`.

Please feel free to add your thoughts, please don't start any implementation as I think we should first discuss, and I am also keen on trying my luck :) 

https://github.com/astral-sh/uv

beeme1mr:
I'm in favor of this. UV seems to be addressing most of the challenges I've encountered while setting up a project.

`ty` also sounds interesting but we should probably hold off on that until it's a bit more mature.

moredip:
I'm a fan of `uv` 👍

lukas-reining:
We also recently started several projects using uv and it was really good so far. Just a really good out-of-the-box experience.
So I am in favor of switching if someone wants to do it :)

aepfli:
as my time is limited, i will unassign me, and open this for everyone to explore/adapt.

gruebel:
finished via #512

PR #512, opened 2025-06-22
feat: starting migration to uv

leohoare:
## This PR
This PR migrates hatch to UV (keeping the build system as hatchling)

### Related Issues
https://github.com/open-feature/python-sdk/issues/502

### Notes / Questions
* Previously scripts were nicely run through hatch. How would we like to keep this functionality now we're migrating to UV?
* Is this project okay with using `astral-sh/setup-uv` in github actions?

### TODOs

- [x] Handle pre-commit
- [x] Sast
- [x] Finalise developer patterns / guides


CONTRIBUTING.md
```diff
@@ -12,20 +12,66 @@ Python 3.9 and above are supported by the SDK.
...
+Run e2e tests with behave:
+```bash
+# Update git submodules and run behave tests
+git submodule update --init --recursive
+cp spec/specification/assets/gherkin/* tests/features/
+uv run behave tests/features/
+rm tests/features/*.feature
+```
+
+### Code Quality
```
aepfli, line 61: i think it would be worth adding how to install the pre-commit hook for ruff too

.github/workflows/release.yml
```diff
@@ -44,18 +44,16 @@ jobs:
...
-        run: pip install --upgrade pip
-
-      - name: Install hatch
-        run: pip install hatch
+      - name: Install dependencies
+        run: uv sync
 
       - name: Build a binary wheel and a source tarball
-        run: hatch build
+        run: uv run python -m build
```
aepfli, line 56:
```suggestion
        run: uv build
```

CONTRIBUTING.md
```diff
@@ -12,20 +12,66 @@ Python 3.9 and above are supported by the SDK.
...
+```
+
+### End-to-End Tests
+
+Run e2e tests with behave:
+```bash
+# Update git submodules and run behave tests
+git submodule update --init --recursive
+cp spec/specification/assets/gherkin/* tests/features/
+uv run behave tests/features/
```
aepfli, line 57: can we maybe configure this in th pyproject.toml, so we only need to execute `uv run behave`

aepfli (commented):
Thank you! I am really looking forward to this. Especially if https://github.com/astral-sh/uv/issues/6365 gets resolved we can add additional targets. Generally i am pretty happy with this starting point. We should focus on ease of use for contributors and IMHO make a lot of opinionated configurations upfront, to reduce the needed knowledge for adopters

CONTRIBUTING.md
leohoare: Yeah, I've been looking for this kind of functionality.
Currently it doesn't look like there's a decent workaround.
See this thread: https://github.com/astral-sh/uv/issues/5903

They're adding support, but the issue has been there for almost a year.

leohoare: Using `project.scripts` is the closest thing I can think of without adding dependencies or other tools.
It'd really be an interim solution...

scripts/scripts.py
```diff
@@ -0,0 +1,38 @@
+# ruff: noqa: S602, S607
+import subprocess
+
+
+def test():
+    """Run pytest tests."""
+    subprocess.run("pytest tests", shell=True, check=True)
```
aepfli, line 7: [suggestion/question] Will this run via UV or directly as Python? This means we would need to have all the packages available on the path for the users, which might be tricky or unexpected. We should ensure that we use UV to run those scripts.

leohoare: Maybe this should be made more clear in the readme.

CONTRIBUTING.md
leohoare: Pre-commit would be installed by `uv sync --extra dev` and the first run `uv run precommit` automatically installs the hooks.
Do you mean think we need to add anything else?

gruebel (approved):
thanks 🍻

PR #523, opened 2025-07-18
fix: merge transaction context into hook context evaluation context (#521)

### LINKED ISSUE #521: Transaction context should be merged into the evaluation context of the hooks
Hi,
During flag evaluation the transaction context is merged into the evaluation context here: https://github.com/open-feature/python-sdk/blob/288bd6bb34f2d5e857ae5b2b17f02f79276049c5/openfeature/client.py#L778
However the `hook_context.evaluation_context` is created before this, so it doesn't contain the attributes or targeting key that exists in the propagated transaction context. This is unexpected. I believe that the `_before_hooks_and_merge_context` should be called inside `_establish_hooks_and_provider` so that the `evaluation_context` field in the `hook_context` can be up to date with all the attributes.

kouk:
## merge transaction context into hook context evaluation context

Currently during flag evaluation the transaction context is merged into the evaluation context here: https://github.com/open-feature/python-sdk/blob/288bd6bb34f2d5e857ae5b2b17f02f79276049c5/openfeature/client.py#L778
However the `hook_context.evaluation_context` is created before this, so it doesn't contain the attributes or targeting key that exists in the propagated transaction context. This is unexpected. It is also different than what [the JS implementation does.](https://github.com/open-feature/js-sdk/blob/d8bd93b6d5256445d12185d639e1cd91800a8a16/packages/server/src/client/internal/open-feature-client.ts#L277)

In this PR:
1. The transaction context is merged with the evaluation context  in `_establish_hooks_and_provider` before creating the hook context. 
2. The `_before_hooks_and_merge_context` method is renamed as `_run_before_hooks_and_update_context` to better reflect its updated functionality which is now restricted to just adding the before hooks results to the context.
3. Renamed a variable in `_run_before_hooks_and_update_context` which was inaptly named IMHO. The result of the before hooks is not the "invocation context".
4. Adds a test to verify the transaction context is properly merged into the hook context's evaluation context

This ensures that hooks have access to the complete evaluation context, including any attributes or targeting key from the transaction context.

### Related Issues

Fixes #521 

### Notes

Currently, both the Python sdk and the JS sdk do not update the hook context's `evaluation_context` field after running the before hooks. This is not explicitly required by the spec here: https://openfeature.dev/specification/sections/evaluation-context/#requirement-323
or here:
https://openfeature.dev/specification/sections/hooks

However it seems to me to be a safe assumption that if the before hooks update the evaluation context then the hook context should reference this updated evaluation context instead of referencing the evaluation context as it was before the before hooks ran. I can open a different issue about this

### DIFF
--- a/tests/test_transaction_context_in_hooks.py
+++ b/tests/test_transaction_context_in_hooks.py
@@ -1,0 +1,51 @@
+from openfeature.api import (
+    set_provider,
+    set_transaction_context,
+    set_transaction_context_propagator,
+)
+from openfeature.client import OpenFeatureClient
+from openfeature.evaluation_context import EvaluationContext
+from openfeature.hook import Hook
+from openfeature.provider.no_op_provider import NoOpProvider
+from openfeature.transaction_context import ContextVarsTransactionContextPropagator
+
+
+class TransactionContextHook(Hook):
+    def __init__(self):
+        self.before_called = False
+        self.transaction_attr_value = None
+
+    def before(self, hook_context, hints):
+        self.before_called = True
+        # Check if the transaction context attribute is in the hook context
+        if "transaction_attr" in hook_context.evaluation_context.attributes:
+            self.transaction_attr_value = hook_context.evaluation_context.attributes[
+                "transaction_attr"
+            ]
+        return None
+
+
+def test_transaction_context_merged_into_hook_context():
+    """Test that transaction context is merged into the hook context's evaluation context."""
+    set_transaction_context_propagator(ContextVarsTransactionContextPropagator())
+
+    provider = NoOpProvider()
+    set_provider(provider)
+
+    client = OpenFeatureClient(domain=None, version=None)
+
+    hook = TransactionContextHook()
+    client.add_hooks([hook])
+
+    transaction_context = EvaluationContext(
+        targeting_key="transaction",
+        attributes={"transaction_attr": "transaction_value"},
+    )
+    set_transaction_context(transaction_context)
+
+    client.get_boolean_value(flag_key="test-flag", default_value=False)
+
+    assert hook.before_called, "Hook's before method was not called"
+    assert hook.transaction_attr_value == "transaction_value", (
+        "Transaction context attribute was not found in hook context"
+    )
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):

kouk:
@beeme1mr AFAICT the python sdk does not do what the JS sdk does. Here's the relevant part of the JS sdk: https://github.com/open-feature/js-sdk/blob/d8bd93b6d5256445d12185d639e1cd91800a8a16/packages/server/src/client/internal/open-feature-client.ts#L329
It apparently updates the evaluation context that is referenced by the hook context (might be misunderstanding the code here so check my assumption).

OTOH the Python SDK does not do this: https://github.com/open-feature/python-sdk/blob/288bd6bb34f2d5e857ae5b2b17f02f79276049c5/openfeature/hook/_hook_support.py#L51

This PR does not currently change this, so there is a discrepancy IMHO between the JS and Python SDKs in this regard. I could update this PR or create a new one to address this discrepancy.

leohoare:
Changes look okay from my end, please address the [linting](https://github.com/open-feature/python-sdk/actions/runs/16375335592/job/46283298755).

> However it seems to me to be a safe assumption that if the before hooks update the evaluation context then the hook context should reference this updated evaluation context instead of referencing the evaluation context as it was before the before hooks ran. I can open a different issue about this

From a consumer, that behaviour is expected but it should be decided by the spec.
We could always raise an issue and take it to openfeature?

tests/test_transaction_context_in_hooks.py
```diff
@@ -0,0 +1,58 @@
...
+        if "transaction_attr" in hook_context.evaluation_context.attributes:
+            self.transaction_attr_value = hook_context.evaluation_context.attributes[
+                "transaction_attr"
+            ]
+        return None
+
+
+def test_transaction_context_merged_into_hook_context():
+    """Test that transaction context is merged into the hook context's evaluation context."""
+    # Set up transaction context propagator
```
gruebel, line 30: can you remove all the inline comments here, they don't add any additional value here

gruebel (commented):
thanks, nice work 🍻  left a small comment

kouk:
Opened two issues in the spec repo:
https://github.com/open-feature/spec/issues/328
https://github.com/open-feature/spec/issues/329

tests/test_transaction_context_in_hooks.py
kouk: Done

aepfli (approved):

thomaspoignant (approved):

PR #527, opened 2025-07-24
chore: switch build backend to uv

gruebel:
## This PR

- switched build backend from hatch to uv
- pinned the uv version to 0.8.x to be better control updates
- enabled uv.lock updates via renovate
- some minor cleanup

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.12.4
     hooks:
-      - id: ruff
+      - id: ruff-check
         args: [--fix]
       - id: ruff-format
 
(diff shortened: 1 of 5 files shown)


beeme1mr:
Hey @gruebel, thanks for the PR. I've heard good things about uv. It's a silly question, but what's the best way to manually test this? 😰

gruebel:
@beeme1mr depending on what you want to test exactly you can build it locally.
```
uv build
```
this creates a wheel and a sdist folder under `dist` and this can be used to install in a separate virtual environment or you could replace it in of our contrib projects. they are still using `hatch` so you would need to adjust the `dependencies` block in the `pyproject.toml`
```
dependencies = [
  "openfeature-sdk @ file:///[path to the repo]/dist/openfeature_sdk-0.8.1-py3-none-any.whl",
  ...
]
```
and then you can run in that subproject folder `hatch test`, if you already had a hatch env setup, then you should uninstall it first
```
hatch env remove default
```

beeme1mr (approved):
Thanks!

PR #548, opened 2025-10-31
chore: make the license field PEP639 compatible

gruebel:
## This PR

- found this out by accident and `uv build` supports it. more info can be found [here](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files)


federicobond (approved):
Good catch!

PR #549, opened 2025-11-14
chore: update license copyright and add cursor to gitignore

jonathannorris:
- Updates LICENSE with correct copyright
- adds `.cursor/` to `.gitignore`


lukas-reining:
The license issue can be ignored. It was a false positive and I added an ignore in FOSSA.

PR #560, opened 2026-01-12
fix: Improved handling within ProviderRegistry

keelerm84:
## This PR

Introduces the following changes to the `ProviderRegistry`.

1. There exists logic to prevent initializing and shutting down providers if they are registered multiple times. However, this logic doesn't account for a provider being registered as both a default, and a domain specific version. There are updates to address both of these situations.
2. A small refactor to change the way provider status clean up is handled.
3. Introduced unit tests to cover registry directly.

All of this functionality is in this single PR, but this can be split into individual PRs if desired. Each commit contains a singular bit of functionality or change, so reviewing individually might be easier.

Assuming we keep the PR as one, I have provided the changelog override necessary for release please.

BEGIN_COMMIT_OVERRIDE
refactor: Delete provider status instead of marking as NOT_READY
fix: Prevent providers from being shutdown multiple times
fix: Prevent providers from being initialized multiple times
END_COMMIT_OVERRIDE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,8 +31,8 @@ Homepage = "https://github.com/open-feature/python-sdk"
 dev = [
     "behave",
     "coverage[toml]>=6.5",
-    "pytest",
-    "pytest-asyncio",
+    "pytest>=9.0.0",
+    "pytest-asyncio>=1.3.0",
     "pre-commit"
 ]
 
@@ -57,7 +57,8 @@ pretty = true
 strict = true
 disallow_any_generics = false
 
-[tool.pytest.ini_options]
+[tool.pytest]
+strict = true
 asyncio_default_fixture_loop_scope = "function"
 
 [tool.coverage.report]
--- a/uv.lock
+++ b/uv.lock
@@ -31,11 +31,11 @@ wheels = [
 
 [[package]]
 name = "cfgv"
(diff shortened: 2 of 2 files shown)


openfeature/provider/_registry.py
```diff
@@ -24,13 +24,18 @@ def __init__(self) -> None:
...
+        if domain is None:
+            raise GeneralError(error_message="No domain")
         providers = self._providers
         if domain in providers:
             old_provider = providers[domain]
             del providers[domain]
-            if old_provider not in providers.values():
+            if (
+                old_provider not in providers.values()
+                and old_provider != self._default_provider
```
gruebel: can you switch the statements, it is faster to check the second one, so if this is false there is not need to check all the providers.

openfeature/provider/_registry.py
```diff
@@ -24,13 +24,18 @@ def __init__(self) -> None:
...
+            ):
                 self._shutdown_provider(old_provider)
-        if provider not in providers.values():
+        if provider not in providers.values() and provider != self._default_provider:
```
gruebel: saem here

keelerm84:
@federicobond first commit as a PR is now open at https://github.com/open-feature/python-sdk/pull/561

PR #580, opened 2026-03-21
chore: replace pre-commit with prek

gruebel:
## This PR

- more and more projects already migrated to `prek` like `cpython`, `airflow` and `ruff`
- uses `uv` instead of `virtualenv` to setup Python hooks, which fits nicely in our setup
- supports parallelization of hooks via `priority` groups


federicobond:
Do we need support for updating hooks with Renovate or is it ok to rely on the existing `pre-commit` support?

PR #565, opened 2026-02-21
feat: add set_evaluation_context method to OpenFeatureClient

vikasrao23:
## Summary

Fixes #500

Implements a setter method for client-level evaluation context, allowing users to set a client-wide evaluation context that is properly merged during flag evaluation.

## Changes

- **openfeature/client.py**: Added `set_evaluation_context(context)` method to `OpenFeatureClient` class
- **tests/test_client.py**: Added 3 comprehensive tests:
  1. `test_client_set_evaluation_context` - Verifies the setter works correctly
  2. `test_client_set_evaluation_context_updates_existing` - Verifies updating an existing context
  3. `test_client_set_evaluation_context_is_merged_during_evaluation` - Verifies proper context merging during flag evaluation

## How it works

The client-level context is merged during flag evaluation with the following precedence (lowest to highest):

1. **API-level context** (global, via `api.set_evaluation_context()`)
2. **Transaction context** (via `api.set_transaction_context()`)
3. **Client context** ← **NEW: set via `client.set_evaluation_context()`**
4. **Invocation context** (per flag evaluation call)
5. **Before hooks context** (injected by hooks)

This matches the behavior described in the [OpenFeature spec](https://github.com/open-feature/spec/blob/main/specification/assets/gherkin/contextMerging.feature).

## Example Usage

```python
from openfeature import api
from openfeature.evaluation_context import EvaluationContext

# Get a client
client = api.get_client()

# Set client-level context
client.set_evaluation_context(EvaluationContext(
    targeting_key="user-123",
    attributes={
        "email": "user@example.com",
        "plan": "premium"
    }
))

# This context will be included in all flag evaluations from this client
value = client.get_boolean_value("feature-flag", False)
```

## Testing

All existing tests pass, and new tests verify:
- The setter correctly updates the client's context
- Client context is properly merged with API, transaction, and invocation contexts
- Context precedence is maintained (invocation > client > transaction > API)

Signed-off-by: vikasrao23 <vikasrao23@users.noreply.github.com>

PR #586, opened 2026-04-11
feat: add client-level evaluation context setter

### LINKED ISSUE #500: [FEATURE] client evaluation context support

buildingisfun23:
## Summary
- Adds `set_evaluation_context()` and `get_evaluation_context()` methods to `OpenFeatureClient`
- Context merging was already implemented correctly (API -> transaction -> client -> invocation), the missing piece was just the setter
- Adds 6 tests covering setter behavior, context replacement, and merge chain for both sync and async evaluations

Fixes #500

### DIFF
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -97,6 +97,12 @@ class OpenFeatureClient:
     def get_metadata(self) -> ClientMetadata:
         return ClientMetadata(domain=self.domain)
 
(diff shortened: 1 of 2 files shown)


gruebel:
please check this comment https://github.com/open-feature/python-sdk/pull/565#discussion_r2843095601 for a previous PR targeting the same issue

PR #589, opened 2026-04-12
fix: correctly reset api state on shutdown

### LINKED ISSUE #585: shutdown() doesn't fully reset state
`api.shutdown()` currently just calls `provider_registry.shutdown()`, which shuts down the providers but doesn't even reset them back to no-op. Hooks, event handlers, evaluation context, transaction context propagators, and provider references are all left in place.

`clear_providers()` does a bit more (resets providers to no-op, clears event handlers), but it also doesn't touch hooks, evaluation context, or transaction context propagators.

Most other SDKs do a full reset on shutdown: shut down providers, reset to no-op, clear hooks, clear event handlers, clear evaluation context, and reset transaction context propagators. We should do the same.

Spec 1.6.2 says the API's shutdown MUST reset all state (see open-feature/spec#375 for a pending clarification to explicitly list evaluation context and transaction context propagators).

Related to 1.0 readiness: #522

gruebel:
## This PR

- added extra `clear` functions for easier reuse

### Related Issues

Fixes #585 

### Notes

@toddbaert should we handle this as "breaking" change?


beeme1mr (approved):

toddbaert (approved):
I don't think I would consider this breaking. It's only behaviorally differently, and in a way I think is quite unlikely to negatively impact anyone.

Will merge next week unless I hear objections.

federicobond (approved):

PR #590, opened 2026-04-14
feat(client): add get/set_evaluation_context methods

### LINKED ISSUE #500: [FEATURE] client evaluation context support

balgaly:
## Summary

Closes #500

Adds `set_evaluation_context()` and `get_evaluation_context()` to `OpenFeatureClient` per the OpenFeature spec requirement 3.2.2.

The client-level context is already merged correctly during flag evaluation (`global → transaction → client → per-call`). These methods expose a public setter/getter so callers can update the context after construction — matching what the Go, Java, and .NET SDKs already provide.

## Changes

- `openfeature/client.py`: add `get_evaluation_context()` and `set_evaluation_context()` with type validation
- `tests/test_client.py`: 4 new tests covering set/get, type error, and integration with flag evaluation

## Tests

```
test_client_set_evaluation_context_updates_context
test_client_get_evaluation_context_returns_current_context
test_client_set_evaluation_context_invalid_type_raises
test_client_set_evaluation_context_is_used_in_flag_evaluation
```

### DIFF
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -100,6 +100,23 @@ class OpenFeatureClient:
     def add_hooks(self, hooks: list[Hook]) -> None:
         self.hooks = self.hooks + hooks
 
+    def get_evaluation_context(self) -> EvaluationContext:
+        """Return the client-level evaluation context."""
+        return self.context
+
+    def set_evaluation_context(self, evaluation_context: EvaluationContext) -> None:
+        """Set a client-level evaluation context.
+
+        The client context is merged after the global and transaction contexts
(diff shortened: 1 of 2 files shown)


balgaly:
Picked up on @gruebel's [note in #565](https://github.com/open-feature/python-sdk/pull/565#discussion_r2843095601) that the actual fix for #500 is the missing E2E tests, not the setter. Going to close this PR and open a new one targeting those tests instead.

balgaly:
Superseded by [#593](https://github.com/open-feature/python-sdk/pull/593), which implements the E2E tests @gruebel requested instead of adding a new setter.

PR #593, opened 2026-04-17
test(e2e): add Behave step definitions for context merging

### LINKED ISSUE #500: [FEATURE] client evaluation context support

balgaly:
Fixes #500

Per @gruebel's [note on #565](https://github.com/open-feature/python-sdk/pull/565#discussion_r2843095601) — the actual fix for #500 is the missing E2E tests, not a new setter, since Python idiomatically uses direct attribute access. The merging precedence in the SDK is already correct (openfeature/client.py:422-429); this PR proves it with Gherkin scenarios from the spec.

## Summary

- Bump the spec submodule to 130df3eb so contextMerging.feature is picked up by the existing `poe e2e` task.
- Add tests/features/environment.py with a before_scenario hook that resets provider, hook, API context, and transaction context state. This prevents cross-scenario state leaks — without it, contextMerging scenarios can pollute the following evaluation.feature run.
- Add tests/features/steps/context_merging_steps.py:
  - RetrievableContextProvider: captures the merged EvaluationContext it receives so assertions can inspect what the SDK passed through.
  - Step definitions covering every scenario in contextMerging.feature: single-level insert, multi-level insert, and per-key overwrite precedence across API / Transaction / Client / Invocation / Before Hooks.

Client-level context is set via direct attribute assignment on OpenFeatureClient.context rather than a new setter. This matches the Pythonic pattern called out in #565 review and requires no API surface change.

## Test plan

- [x] poe e2e (with PYTHONIOENCODING=utf-8 on Windows): 4 features / 50 scenarios / 233 steps passing
- [x] poe test: 155 pytest tests passing
- [x] ruff check + ruff format clean

## Supersedes

Closes #590 in favor of this approach.


tests/features/steps/context_merging_steps.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def _ensure_state(context: typing.Any) -> None:
+    if getattr(context, "_merging_initialized", False):
+        return
+    api.clear_providers()
+    api.clear_hooks()
+    api.set_evaluation_context(EvaluationContext())
+    set_transaction_context_propagator(ContextVarsTransactionContextPropagator())
+    set_transaction_context(EvaluationContext())
```
gruebel: i think this can be replaced by `api.shutdown()` now. please rebase or merge main branch, then you will see the new behavior.

tests/features/steps/context_merging_steps.py
```diff
@@ -0,0 +1,215 @@
...
+@given("a stable provider with retrievable context is registered")
+def step_impl_retrievable_provider(context: typing.Any) -> None:
+    context._merging_initialized = False
+    _ensure_state(context)
+
+
+def _add_entry_at_level(
+    context: typing.Any, level: str, key: str, value: typing.Any
+) -> None:
+    _ensure_state(context)
```
gruebel: why is `_ensure_state()` invoked in multiple functions?

tests/features/steps/context_merging_steps.py
```diff
@@ -0,0 +1,215 @@
...
+def step_impl_evaluate(context: typing.Any) -> None:
+    _ensure_state(context)
+    hook = _BeforeHookContextInjector(context.before_hook_context)
+    context.client.add_hooks([hook])
+    try:
+        context.client.get_boolean_details(
+            "some-flag", False, context.invocation_context
+        )
+    finally:
+        context.client.hooks = [h for h in context.client.hooks if h is not hook]
```
gruebel: why do we need to do this check here?

tests/features/steps/context_merging_steps.py
```diff
@@ -0,0 +1,215 @@
...
+        context.invocation_context = EvaluationContext(
+            targeting_key=current.targeting_key,
+            attributes={**current.attributes, key: value},
+        )
+    elif level == "Before Hooks":
+        current = context.before_hook_context
+        context.before_hook_context = EvaluationContext(
+            targeting_key=current.targeting_key,
+            attributes={**current.attributes, key: value},
+        )
```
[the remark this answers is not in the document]
gruebel: gemini is correct about the `merge()` method usage, no need to do this manually here.

gruebel (commented):
thanks for the work 🍻 overall looks good, but I added a few comments

toddbaert:
Hey @balgaly ... will you be picking this back up? Altnernatively, @karnyadavdev may want to take it over since they opened something similar as a duplicate.

karnyadavdev:
@toddbaert, sounds good. I can take this over and push the remaining fixes

gruebel (approved):

PR #595, opened 2026-05-20
feat!: make set_provider non-blocking, add set_provider_and_wait

### LINKED ISSUE #594: set_provider() should be non-waiting
tl;dr: the default `set_provider()` / recommended way to initialize should be non-blocking. Blocking-by-default may lead to hard-to-debug denial of service.

> [Requirement 1.1.2.4](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1124)
>> The API SHOULD provide functions to set a provider and wait for the initialize function to complete or abnormally terminate.
>
> This function not only sets the provider, but ensures that the provider is ready (or in error) before returning or settling.
>
> ```
> // default provider
> OpenFeatureAPI.getInstance().setProviderAndWait(myprovider); // this method blocks until the provider is ready or in error
> // client uses the default provider
> Client client = OpenFeatureAPI.getInstance().getClient(); 
> 
> // provider associated with domain-1
> OpenFeatureAPI.getInstance().setProviderAndWait('domain-1', myprovider); // this method blocks until the provider is ready or in error
> // client uses provider associated with the domain named 'domain-1'
> Client client = OpenFeatureAPI.getInstance().getClient('domain-1');
> ```

Requirement 1.1.2.4 implies that the default `set_provider()` should not wait for the provider to initialize.

Having the waiting implementation being the default has a couple of issues. The recommended `api.set_provider(MyProvider())` blocks the rest of the application from proceeding. This turns a provider hanging during initialization into an application denial of service.

This can be exaggerated further by something like gunicorn which has an default 30s worker initialization timeout. If the provider takes 30+ seconds to initialize, this leads to gunicorn killing the worker with a cryptic `WORKER TIMEOUT` message which may be hard to debug.

If `set_provider()` is non-waiting, the worst case scenario is evaluations returning default values which is much safer. For users who do want to wait, we should provide `set_provider_and_wait()`

jonathannorris:
## Summary

- `set_provider()` now returns immediately; provider initialization runs in a background thread (daemon)
- Adds `set_provider_and_wait()` that blocks until initialization completes or re-raises on failure
- Updates existing tests to use `set_provider_and_wait` where init completion is required; adds 8 new tests covering non-blocking semantics explicitly

## Motivation

Closes #594. Spec [Requirement 1.1.2.4](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1124) implies `set_provider()` should be non-blocking, with a separate variant for callers who need to wait. Python was the only SDK where `set_provider()` blocked by default. The gunicorn DoS scenario in the issue is real — a provider that hangs during init causes gunicorn to kill the worker with a cryptic `WORKER TIMEOUT`.

All other SDKs (Go, Java, JS/TS, Ruby) have the same split: a fire-and-forget default and a `*AndWait` variant. This implementation follows Ruby's approach since it's also synchronous — a `wait_for_init` flag in the registry routes to either `Thread(daemon=True).start()` or an inline call.

## Behavior change

`set_provider()` — non-blocking (default):
- Returns before `initialize()` completes
- Flag evaluations during the init window return the default value with `PROVIDER_NOT_READY`
- Initialization errors fire a `PROVIDER_ERROR` event; they are not propagated to the caller

`set_provider_and_wait()` — blocking:
- Blocks until `initialize()` completes or raises
- Provider is `READY` (or in error state) before the call returns
- Re-raises the original exception on failure

### Notes

This is a breaking behavioral change for callers who relied on `set_provider()` blocking until ready. Tagging as `feat!` so release-please bumps the minor version (pre-1.0, so 0.9.x → 0.10.0 per `bump-minor-pre-major`). Getting this in before 1.0 is the right call to avoid carrying it as a breaking change post-stable.

Relates to: https://github.com/open-feature/python-sdk/issues/96
Fixes: https://github.com/open-feature/python-sdk/issues/594


jonathannorris:
There's a related PR (#567) worth flagging. It takes a purely additive approach — `set_provider()` stays blocking, and `set_provider_and_wait()` is added as a second blocking method. That doesn't fix the DoS scenario from #594; a hanging provider still blocks the caller.

This PR makes `set_provider()` actually non-blocking (background thread), matching Go, Java, JS/TS, and Ruby. It's a breaking behavioral change, which is exactly why the `1.0-release` label matters.

openfeature/provider/_registry.py
```diff
@@ -36,15 +40,17 @@ def set_provider(self, domain: str, provider: FeatureProvider) -> None:
             ):
                 self._shutdown_provider(old_provider)
```
dd-oleksii: major: `_shutdown_provider` call remains blocking and may cause a similar issue

openfeature/provider/_registry.py
dd-oleksii: major: similar here — shutdown is blocking regardless of the flag.

openfeature/provider/_registry.py
```diff
@@ -75,8 +81,24 @@ def shutdown(self) -> None:
...
+                daemon=True,
+            )
+            thread.start()
+
+    def _run_initialize(
+        self, provider: FeatureProvider, raise_on_error: bool = False
+    ) -> None:
         try:
             if hasattr(provider, "initialize"):
                 provider.initialize(self._get_evaluation_context())
```
dd-oleksii: minor: this is where we call user-defined code — I think it's a better place to spawn thread. i.e. we don't need a thread if `provider` doesn't have an `initialize` method

dd-oleksii (approved):

toddbaert (approved):
Pulled locally and added a few fixes:

- closed the race where concurrent `set_provider` calls with the same instance could double-initialize
- prevented a late `PROVIDER_READY/PROVIDER_ERROR` from a background init from clobbering the status of a provider that was already replaced or shut down_
- eliminated a `KeyError` in shutdownby using `pop` not `del`
- made old-provider shutdown async so a hanging shutdown() can't block set_provider (otherwise the DoS this PR fixes is only half-closed).

I added tests that seemed reasonable for these. Python is not my strong-suit so please take a look @jonathannorris .

commit is [here](https://github.com/open-feature/python-sdk/pull/595/changes/a5988bdfbe617dad899e939ec02ccb521c2780c8).

I think this resolves all the review feedback by bots and @dd-oleksii .

toddbaert:
Leaving this open for a bit because it's a pretty fundamental change, but I will get it merged this week. Please weigh in if you have objections.

jonathannorris:
We found an issue with the stale check @toddbaert added in `_run_initialize`. It checks `provider not in self._provider_status`, but replaced providers stay in `_provider_status` until `_shutdown_provider` pops them — and since shutdown runs asynchronously via `_shutdown_if_unused`, there's a window where a stale provider's `initialize()` can complete, pass the check, and dispatch `PROVIDER_READY` even though it's no longer bound to any domain.

The fix is to check active registration instead:

```python
# current (insufficient)
if provider not in self._provider_status:
    return

# correct
if provider is not self._default_provider and provider not in self._providers.values():
    return
```

Will push a fix.

leohoare (commented):
It looks like there is a race condition on shutdown.

Now that old-provider shutdown runs in the background, re-registering the same provider instance while its previous shutdown is still finishing lets the stale shutdown wipe out the fresh registration's event wiring and status.

Proposed Fix (or something like):
* Add a per-registration generation token (a counter bumped each time a provider is registered).
* Have each background shutdown capture the token it was started for.
* Before the shutdown does its final `detach()` and status pop, re-check that the provider's current token still matches.
* If the provider has been re-registered since (newer token), skip the cleanup instead of clobbering the new setup.

toddbaert:
> It looks like there is a race condition on shutdown.
> 
> Now that old-provider shutdown runs in the background, re-registering the same provider instance while its previous shutdown is still finishing lets the stale shutdown wipe out the fresh registration's event wiring and status.
> 
> Proposed Fix (or something like):
> 
> * Add a per-registration generation token (a counter bumped each time a provider is registered).
> * Have each background shutdown capture the token it was started for.
> * Before the shutdown does its final `detach()` and status pop, re-check that the provider's current token still matches.
> * If the provider has been re-registered since (newer token), skip the cleanup instead of clobbering the new setup.

Good call. This race is valid; repro added as `test_stale_shutdown_does_not_clobber_re_registered_provider`.
My simpler fix: pass a flag from `_shutdown_if_unused` so `_shutdown_provider` re-checks under the lock _after_ `provider.shutdown()` returns; skip pop/detach if re-registered. Sync `shutdown()` path unaffected. This makes the test pass and fixes the bug, I think... see: [commit](https://github.com/open-feature/python-sdk/pull/595/changes/c72d70538cdfa87113a66c225f9bf5110aca1ce2).

PR #598, opened 2026-05-27
fix: dispatch event handlers asynchronously

### LINKED ISSUE #596: Event handler error isolation and async execution
Several issues with event handler invocation in `_event_support.py`:

**1. No error isolation (spec 5.2.5 violation)**

Handler calls (lines 33, 39, 101) have no `try`/`except`. If a handler raises, subsequent handlers won't run.

Spec 5.2.5:
> If a `handler function` terminates abnormally, other `handler functions` **MUST** run.

**2. Handlers block the emitting thread**

Handlers run synchronously on the calling thread, and under a lock. A slow handler blocks all event operations. Other SDKs dispatch handlers asynchronously:

- [Java](https://github.com/open-feature/java-sdk/blob/0c308968826f2a45fed492b115ecbb84a17f0765/src/main/java/dev/openfeature/sdk/EventSupport.java#L129-L137) submits each handler to a `ThreadPoolExecutor`
- [Go](https://github.com/open-feature/go-sdk/blob/4a96979d40ad08eb824a711bac01ed907a28322a/openfeature/event_executor.go#L348-L365) launches a goroutine per handler
- [.NET](https://github.com/open-feature/dotnet-sdk/blob/29fd9c167a41cee1099aff9618d40b24ef9bdce6/src/OpenFeature/EventExecutor.cs#L239-L260) processes events on a background thread via `Task.Run` with a `Channel`

Handlers should be dispatched via a `ThreadPoolExecutor` so they don't block the emitter.

**3. Potential deadlocks**

Because handlers execute while holding an `RLock`, a provider or handler that interacts with the event system from a different thread during handling could deadlock. Dispatching handlers outside the lock (e.g., via a `ThreadPoolExecutor`) would eliminate this risk.

**Testing**

There are currently no tests covering handler error isolation. Tests should be added to verify that a raising handler does not prevent subsequent handlers from running.

iccccccccccccc:
Fixes #596

## Changes
- dispatch global, client, and immediate event handlers through a shared thread pool
- copy handler lists while holding the registry locks, then run handlers outside those locks
- isolate handler exceptions so one failing handler doesn't stop the rest
- add regression coverage for handler error isolation and non-blocking event emission

## Verification
- `uv run --frozen pytest tests -q`
- `uv run --frozen ruff check`
- `uv run --frozen ruff format --check openfeature/_event_support.py tests/test_api.py tests/test_client.py`
- `git diff --check`

### DIFF
--- a/openfeature/_event_support.py
+++ b/openfeature/_event_support.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import threading
 import typing
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
+from logging import getLogger
 
 from openfeature.event import (
     EventDetails,
@@ -16,6 +18,8 @@ if typing.TYPE_CHECKING:
     from openfeature.client import OpenFeatureClient
 
 
+_logger = getLogger(__name__)
+
 _global_lock = threading.RLock()
 _global_handlers: dict[ProviderEvent, list[EventHandler]] = defaultdict(list)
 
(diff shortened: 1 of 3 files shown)


toddbaert:
Closing as a duplicate of #599, which addresses the full scope of #596 (both error isolation and async dispatch). Thanks for the contribution.

PR #599, opened 2026-05-27
fix: isolate provider event handler dispatch

### LINKED ISSUE #596: Event handler error isolation and async execution

Lucas-FManager:
## Problem

Provider event handlers currently execute synchronously while the handler registry lock is held. That means a handler exception can stop later handlers, and a slow handler can block the provider/client path that emitted the event.

## Approach

- Snapshot global/client handlers under the lock, then release the lock before dispatch.
- Dispatch handlers through a shared `ThreadPoolExecutor`.
- Wrap each handler invocation so one failing handler is logged and does not prevent the next handler from running.
- Avoid the nested client-handler lock pattern in provider dispatch by snapshotting matching clients first.
- Update event tests to wait for async handler dispatch and add coverage for error isolation and non-blocking emission.

## Security

This avoids running arbitrary user event-handler code while internal handler registries are locked. It also keeps handler exceptions contained to the callback boundary instead of propagating into SDK/provider control flow.

## Verification

- `uv run pytest tests -q` -> 158 passed
- `uvx ruff check openfeature/_event_support.py tests/test_client.py tests/test_api.py` -> passed
- `uvx ruff format --check openfeature/_event_support.py tests/test_client.py tests/test_api.py` -> passed
- `git diff --check` -> passed

Fixes #596

### DIFF
--- a/openfeature/_event_support.py
+++ b/openfeature/_event_support.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import threading
 import typing
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
+from logging import getLogger
 
 from openfeature.event import (
     EventDetails,
@@ -15,6 +17,9 @@ from openfeature.provider import FeatureProvider, ProviderStatus
 if typing.TYPE_CHECKING:
     from openfeature.client import OpenFeatureClient
 
+
+_logger = getLogger(__name__)
+_event_executor = ThreadPoolExecutor(thread_name_prefix="openfeature-event-handler")
 
 _global_lock = threading.RLock()
 _global_handlers: dict[ProviderEvent, list[EventHandler]] = defaultdict(list)
@@ -29,14 +34,22 @@ def run_client_handlers(
     client: OpenFeatureClient, event: ProviderEvent, details: EventDetails
 ) -> None:
     with _client_lock:
-        for handler in _client_handlers[client][event]:
-            handler(details)
+        handlers_by_event = _client_handlers.get(client)
+        if handlers_by_event is None:
+            return
+
+        handlers = tuple(handlers_by_event.get(event, ()))
+
+    for handler in handlers:
+        _submit_handler(handler, details)
 
 
 def run_global_handlers(event: ProviderEvent, details: EventDetails) -> None:
     with _global_lock:
-        for handler in _global_handlers[event]:
-            handler(details)
+        handlers = tuple(_global_handlers.get(event, ()))
+
+    for handler in handlers:
+        _submit_handler(handler, details)
 
 
 def add_client_handler(
@@ -83,9 +96,12 @@ def run_handlers_for_provider(
     run_global_handlers(event, details)
     # run the handlers for clients associated to this provider
     with _client_lock:
-        for client in _client_handlers:
-            if client.provider == provider:
-                run_client_handlers(client, event, details)
+        clients = tuple(
+            client for client in _client_handlers if client.provider == provider
+        )
+
+    for client in clients:
+        run_client_handlers(client, event, details)
 
 
 def _run_immediate_handler(
@@ -98,7 +114,20 @@ def _run_immediate_handler(
         ProviderStatus.STALE: ProviderEvent.PROVIDER_STALE,
     }
     if event == status_to_event.get(client.get_provider_status()):
-        handler(EventDetails(provider_name=client.provider.get_metadata().name))
+        _submit_handler(
+            handler, EventDetails(provider_name=client.provider.get_metadata().name)
+        )
+
+
+def _submit_handler(handler: EventHandler, details: EventDetails) -> None:
+    _event_executor.submit(_run_handler, handler, details)
+
+
+def _run_handler(handler: EventHandler, details: EventDetails) -> None:
+    try:
+        handler(details)
+    except Exception:
+        _logger.exception("Unhandled exception in OpenFeature event handler")
 
 
 def clear() -> None:
--- a/openfeature/_event_support.py
+++ b/openfeature/_event_support.py
@@ -18,7 +18,7 @@ if typing.TYPE_CHECKING:
     from openfeature.client import OpenFeatureClient
 
 
-_logger = getLogger(__name__)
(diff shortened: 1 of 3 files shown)


Lucas-FManager:
Updated in the latest push: `run_client_handlers()` now uses `_client_handlers.get(client)` and then `handlers_by_event.get(event, ())`, and `run_global_handlers()` uses `_global_handlers.get(event, ())`. That avoids mutating the `defaultdict` during read-only dispatch while preserving the async/error-isolated behavior.

Re-verified after the change:
- `uv run pytest tests -q` -> 158 passed
- `uvx ruff check openfeature/_event_support.py tests/test_client.py tests/test_api.py` -> passed
- `uvx ruff format --check openfeature/_event_support.py tests/test_client.py tests/test_api.py` -> passed
- `git diff --check` -> passed

openfeature/_event_support.py
```diff
@@ -16,6 +18,9 @@
     from openfeature.client import OpenFeatureClient
 
 
+_logger = getLogger(__name__)
```
gruebel:
```suggestion
logger = getLogger("openfeature")
```
`logger` instances are usually not marked as private and we use the same name in different areas.

openfeature/_event_support.py
```diff
@@ -16,6 +18,9 @@
...
+_event_executor = ThreadPoolExecutor(thread_name_prefix="openfeature-event-handler")
```
gruebel, line 23: this creates a default worker count, which can't be adjusted by the user.

@toddbaert do you know how this is handled by other languages? we could expose an env var to control the number, if needed.

tests/test_client.py
```diff
@@ -556,6 +572,67 @@ def test_provider_event_handler_exception():
...
+    details = ProviderEventDetails(error_code=ErrorCode.GENERAL, message="some_error")
+    expected_details = EventDetails.from_provider_event_details(
+        provider.get_metadata().name, details
+    )
+
+    # When
+    provider.emit_provider_error(details)
+
+    # Then
+    assert handler_called.wait(timeout=1)
```
gruebel, line 617: I think you are missing a check for `raising_handler` to be called, right?

gruebel (commented):
overall looks good and addresses the mentioned issues. added just a few comments, thanks 🍻

openfeature/_event_support.py
Lucas-FManager: System.Collections.Hashtable.body

openfeature/_event_support.py
Lucas-FManager: System.Collections.Hashtable.body

tests/test_client.py
Lucas-FManager: System.Collections.Hashtable.body

Lucas-FManager:
Small follow-up after the Codecov report: I added a focused regression test for the read-only client-dispatch no-op path, so `_event_support.py` is locally covered at 100%.

Verification after the new commit:
- `uv run pytest tests/test_client.py tests/test_api.py -q` -> 74 passed
- `uv run coverage run -m pytest tests/test_client.py tests/test_api.py -q; uv run coverage report -m openfeature/_event_support.py` -> `_event_support.py` 100%
- `uvx prek run --all-files --show-diff-on-failure` -> passed

The main build workflow is still waiting for maintainer approval for the fork run, so there is no new CI failure on the latest head yet.

openfeature/_event_support.py
toddbaert: > @toddbaert do you know how this is handled by other languages? we could expose an env var to control the number, if needed.

We have a few other cases where we hardcode some things like this - I think it's fine.

> System.Collections.Hashtable.body

@Lucas-FManager I think this message is automated.

.pre-commit-config.yaml
```diff
@@ -26,6 +26,7 @@ repos:
     hooks:
       - id: pyproject-fmt
         priority: 0
+        additional_dependencies: [toml-fmt-common==1.2.0]
```
toddbaert:
```suggestion
```

toddbaert:
I made some small test improvements, and also shut down the executor on exit.

tests/test_client.py
toddbaert: @gruebel I think this is fixed now.

toddbaert (approved):

tests/test_client.py
gruebel: yeah, looks like 👍

gruebel (approved):

PR #610, opened 2026-05-30
feat: client evaluation context getter and setter

### LINKED ISSUE #500: [FEATURE] client evaluation context support

karnyadavdev:
closes #500

Exposes `evaluation_context` as a property getter and setter on `OpenFeatureClient` to support getting and setting a client-wide evaluation context, matching requirements from the OpenFeature specification.

Added a dedicated unit test in `tests/test_client.py` to verify getting and setting.


karnyadavdev:
review this pr

gruebel:
Duplicate of #593

PR #614, opened 2026-06-19
fix: thread safety issues

leakonvalinka:
## This PR

- aims to improve the thread-safety of the python-sdk
- some of these changes still need to be discussed -> see comments (mentioning potential issues and options)

### Related Issues

https://github.com/open-feature/python-sdk/issues/96

### Notes

- i'm not a python expert, please let me know if i got anything wrong :)


aepfli:
I am curious, can we add a tool to detect this like vmlens in java? In the sense of automated testing?

openfeature/provider/_registry.py
```diff
@@ -54,9 +55,16 @@ def set_provider(
...
     def get_provider(self, domain: str | None) -> FeatureProvider:
-        if domain is None:
-            return self._default_provider
-        return self._providers.get(domain, self._default_provider)
+        # defensive lock under the GIL as the op is basically atomic
+        # but we might want to keep it so a provider that's about
+        # to be shut down isn't returned
+        # however it contributes to a potential deadlock that is currently
+        # still in place (clear_providers: registry's lock -> _event_support's lock;
+        # run_handlers_for_provider: _event_support's lock -> registry's lock)
```
leakonvalinka: just writing a comment here as a TODO that still needs to be resolved before merging

openfeature/hook/__init__.py
```diff
@@ -24,6 +25,7 @@
 ]
 
 _hooks: list[Hook] = []
+_hooks_lock = threading.Lock()
```
gruebel: can you use an `RLock` for consistency and it is also the safer option.

openfeature/client.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         self.version = version
         self.context = context or EvaluationContext()
         self.hooks = hooks or []
+        self._hooks_lock = threading.Lock()
```
gruebel: same here

openfeature/provider/__init__.py
```diff
@@ -261,5 +261,6 @@ def emit_provider_stale(self, details: ProviderEventDetails) -> None:
         self.emit(ProviderEvent.PROVIDER_STALE, details)
 
     def emit(self, event: ProviderEvent, details: ProviderEventDetails) -> None:
-        if hasattr(self, "_on_emit"):
-            self._on_emit(self, event, details)
+        on_emit = getattr(self, "_on_emit", None)
+        if on_emit is not None:
+            on_emit(self, event, details)
```
gruebel, line 266: why is this change needed?

openfeature/provider/_registry.py
gruebel: i don't think we gain anything by adding a lock here, because the shutdown can also happen directly after you get the provider.

openfeature/provider/_registry.py
```diff
@@ -83,7 +91,8 @@ def set_default_provider(
             self._shutdown_if_unused(old_provider)
 
     def get_default_provider(self) -> FeatureProvider:
-        return self._default_provider
+        with self._lock:
```
gruebel: same here, not needed

openfeature/provider/_registry.py
```diff
@@ -93,11 +102,13 @@ def clear_providers(self) -> None:
             self._provider_status = {
                 self._default_provider: ProviderStatus.READY,
             }
+            clear_event_handlers()
```
gruebel: this is already handled in the API level `clear_providers()` function and shouldn't be mixed in here

openfeature/provider/_registry.py
```diff
@@ -93,11 +102,13 @@ def clear_providers(self) -> None:
...
 
     def shutdown(self) -> None:
         with self._lock:
             providers = {self._default_provider, *self._providers.values()}
 
+        # do we want to move this inside the lock? it allows a narrow double-shutdown window
```
gruebel: this was done intentionally to not hold the lock for too long and risk a deadlock

openfeature/provider/_registry.py
```diff
@@ -214,7 +225,12 @@ def _shutdown_provider(
         provider.detach()
 
     def get_provider_status(self, provider: FeatureProvider) -> ProviderStatus:
-        return self._provider_status.get(provider, ProviderStatus.NOT_READY)
+        # defensive lock under the GIL as the op is basically atomic
+        # but we might want to keep it so a provider that's about
+        # to be shut down isn't returned
+        # however, removing it would enable moving _run_immediate_handler into the lock i think
+        with self._lock:
```
gruebel: as mentioned before, not needed.

openfeature/transaction_context/__init__.py
```diff
@@ -21,25 +23,28 @@
...
+        _evaluation_transaction_context_propagator = transaction_context_propagator
 
 
 def clear_transaction_context_propagator() -> None:
     set_transaction_context_propagator(NoOpTransactionContextPropagator())
 
 
 def get_transaction_context() -> EvaluationContext:
-    return _evaluation_transaction_context_propagator.get_transaction_context()
+    with _propagator_lock:
```
gruebel: not sure about this one here, if this is really needed or not

openfeature/api.py
```diff
@@ -62,7 +62,6 @@ def set_provider_and_wait(provider: FeatureProvider, domain: str | None = None)
 
 def clear_providers() -> None:
     provider_registry.clear_providers()
-    _event_support.clear()
```
gruebel, line 65: and this should be added back

gruebel (commented):
thanks 🍻 overall it looks pretty good and adds more thread safety especially when using free-threaded CPython

openfeature/provider/__init__.py
leakonvalinka: because of what was pointed out in the original issue:
"
`AbstractProvider._on_emit` (`provider/__init__.py`) - `emit()` does `if hasattr(self, "_on_emit"): self._on_emit(...)` which is a TOCTOU; `detach()` during shutdown can delete `_on_emit` while a background thread is between the check and the call
"

leakonvalinka:
@aepfli I don't think so, unfortunetly. I did try to write a few tests and also have Claude write some with `threading.Barrier` & run them a few thousand times with `pytest repeat`, but they didn't reliably catch any issues even before the fixes were implemented.

openfeature/transaction_context/__init__.py
leakonvalinka: Same here, I tried to ask Claude but the answer was sometimes yes and sometimes no 😅

openfeature/transaction_context/__init__.py
```diff
@@ -21,25 +23,28 @@
...
 
 
 def set_transaction_context(evaluation_context: EvaluationContext) -> None:
-    global _evaluation_transaction_context_propagator
-    _evaluation_transaction_context_propagator.set_transaction_context(
-        evaluation_context
-    )
+    with _propagator_lock:
+        propagator = _evaluation_transaction_context_propagator
+        propagator.set_transaction_context(evaluation_context)
```
toddbaert: We're holding `_propagator_lock` across the propagator's get and set, but I think only the reference read needs guarding; the swap in `set_transaction_context_propagator` is the sole writer, and reading the global is atomic anyway, if I'm understanding the Python docs right...

Holding it through the call serializes all transaction-context access across threads, and might be problematic especially if somebody was doing a lot in a custom propagator.

```suggestion
def get_transaction_context() -> EvaluationContext:
    with _propagator_lock:
        propagator = _evaluation_transaction_context_propagator
    return propagator.get_transaction_context()


def set_transaction_context(evaluation_context: EvaluationContext) -> None:
    with _propagator_lock:
        propagator = _evaluation_transaction_context_propagator
    propagator.set_transaction_context(evaluation_context)
```

leakonvalinka: I can change it, I'm really not sure myself..

openfeature/hook/__init__.py
```diff
@@ -151,14 +153,21 @@ def supports_flag_value_type(self, flag_type: FlagType) -> bool:
         return True
 
 
+# while the lock guarantees safety, even without it there was never a loss within 50.000 runs (with the default GIL
+# switch interval of 5ms). only when the switch interval was significantly shortened to 0.1 microseconds, losses were
+# observed without locks every now and then. with a no-GIL python, the lock would be essential
```
dd-oleksii: minor: this comment seems trying to justify that it's "fine" to not have a lock in some cases?

We know the bug is here, and the first rule of testings tells us that we can't prove the absence of bugs with tests — only their presence. Longer switch interval doesn't make it safe — just make less likely to be observed. And as one person put it: when you serve millions of users, once-in-a-million events happen every day

Anyway, this comment doesn't look like it belongs to the code — once the lock is here, there's not much point in debating/defending it

openfeature/hook/__init__.py
```diff
@@ -24,6 +25,7 @@
 ]
 
 _hooks: list[Hook] = []
+_hooks_lock = threading.RLock()
```
dd-oleksii, line 28: minor: the critical sections are very short and don't invoke anything else, so we don't need a re-entrant lock here — a normal one would do
```suggestion
_hooks_lock = threading.Lock()
```

openfeature/client.py
```diff
@@ -98,7 +100,10 @@ def get_metadata(self) -> ClientMetadata:
         return ClientMetadata(domain=self.domain)
 
     def add_hooks(self, hooks: list[Hook]) -> None:
-        self.hooks = self.hooks + hooks
+        # Guards the read-concat-store against a lost update; this practically never races under the default 5ms GIL
+        # switch interval, but is essential under a no-GIL build.
```
dd-oleksii: minor: misleading comment again. It's not "never" — just rarely

openfeature/_event_support.py
```diff
@@ -133,7 +141,6 @@ def _run_handler(handler: EventHandler, details: EventDetails) -> None:
 
 
 def clear() -> None:
-    with _global_lock:
+    with _global_lock, _client_lock:
```
dd-oleksii: minor: this increases the length of critical sections, increasing the change of contention. Not that it's called too frequently but it's moving in the opposite direction

openfeature/_event_support.py
```diff
@@ -61,6 +61,13 @@ def add_client_handler(
         handlers = _client_handlers[client][event]
         handlers.append(handler)
 
+    # outside the lock intentionally: the immediate-fire status check acquires the registry lock, so calling it
+    # under _client_lock risks lock-order inversion against run_handlers_for_provider (registry lock → _client_lock).
+    # As a consequence, a narrow double-fire is possible: if dispatch_event(client's event) runs concurrently, it
+    # sets the matching provider status (enabling the immediate fire below) and then re-runs every handler for this
+    # client. If _run_immediate_handler lands after that status set but before dispatch snapshots the handler list,
+    # the handler fires twice — once here, once from dispatch. Only happens when the registered event matches the event
+    # being dispatched; otherwise the immediate fire is a no-op.
```
dd-oleksii: minor: this comment says that we're trading one concurrency issue for another — not a good place to be in. `_run_immediate_handler` is not synchronized with `run_handlers_for_provider`, so a provider event may race with the immediate handler and the user may receive events in reverse order. We'd need to do something about it

openfeature/hook/__init__.py
dd-oleksii: ok, I see @gruebel asked for the `RLock`, so I'm retracting my comment

openfeature/hook/__init__.py
leakonvalinka: Agree, I would also remove this comment before merging the PR. I added it to give people some context and share my findings, as personally I wouldn't have known this by just looking at the changed code. As I am not that well-versed in Python to be able to tell whether this is a basically-impossible issue or how expensive an unnecessary lock would be, I thought maybe someone else could share their knowledge here.

openfeature/_event_support.py
leakonvalinka: Ah I'm sorry, this comment is outdated now that I have removed the locks for the `get` methods in `_registry.py`. However, after taking another look I don't think the double-fire possibility is entirely eliminated either way - even if `_run_immediate_handler` was moved inside the lock here. I will therefore adjust the comment, should I leave the method call outside the lock or move it inside?

openfeature/_event_support.py
leakonvalinka: Fair point, I changed it because it was a mentioned issue in https://github.com/open-feature/python-sdk/issues/96, but I guess not changing it makes more sense..?

openfeature/hook/__init__.py
dd-oleksii: I usually put context comment like this in commit message/PR description and as github comments. Github comments have the highest chance of being noticed by other reviewers, but unfortunately don't get saved in git history 😞

openfeature/_event_support.py
dd-oleksii: We need to serialize subscription and calls to "immediate handler" with respect to events fired by the provider. So updates to client status, firing of immediate handler, and provider event handlers must be synchronized in some way — either a lock or some kind of singe-threaded queue. Not sure if the client lock here is enough — need to do a deeper dive into the code

openfeature/provider/__init__.py
gruebel: gotcha, checked the current code.

openfeature/transaction_context/__init__.py
gruebel: let's keep the lock on the setter for now and remove it from the getter, then it is similar to the other lock implementations. i'm also a bit afraid of holding the lock for too long in the getter.

openfeature/_event_support.py
gruebel: @leakonvalinka let's remove the comment and @dd-oleksii feel free to dive into the code and if you have some good suggestions, then please open up separate issues/PRs. I think this PR is open for long enough 😅

gruebel:
@leakonvalinka let me know, if you need some help to finish up the last parts. I don't want you to be stuck with this PR for much longer, sorry!

leakonvalinka:
@gruebel Thank you! I implemented the remaining feedback, if you notice something else feel free to just change it if you want.

gruebel (approved):
thanks for contribution 🍻 i'll wait till end of week, if there are no other objections

beeme1mr (approved):

PR #616, opened 2026-07-02
feat(core): pass bound domain to provider initialize and enforce domain-scoped binding

### LINKED ISSUE #615: Supply the bound domain to provider initialization
The spec now supplies the bound `domain` to a provider's `initialize` function, and adds an opt-in `domain-scoped` provider declaration that the API enforces by binding such a provider to at most one `domain`. This lets stateful providers (e.g. an OFREP static-context provider with a persistent cache) scope their state to a single, unambiguous `domain`.

### What needs to happen

- Pass the bound `domain` (if any) to the provider's `initialize` when it's registered via the provider mutator
- Extend the provider `initialize` to accept the optional `domain` (non-breaking, additional optional parameter)
- Add a way for a provider to declare itself `domain-scoped`
- Enforce that a `domain-scoped` provider is bound to at most one `domain`, rejecting any attempt to bind an already-bound instance to another `domain`
- Associated testing and docs

### Links

- Parent issue: https://github.com/open-feature/spec/issues/403
- Spec sections: [1.1 Flag evaluation](https://openfeature.dev/specification/sections/flag-evaluation), [2.4 Initialization](https://openfeature.dev/specification/sections/providers)

jonathannorris:
## Summary

- Add optional `domain` parameter to provider `initialize` and pass the bound domain from the provider registry (spec 1.1.2.2, 2.4.1)
- Add `domain_scoped` provider declaration and reject binding a domain-scoped instance to more than one domain (spec 2.4.3, 1.1.8.1)
- Add registry and API tests, including legacy single-argument `initialize` backward compatibility

## Motivation

Unblocks OFREP static-context providers that need the bound `domain` at init time to scope persisted cache keys per [open-feature/spec#393](https://github.com/open-feature/spec/pull/393) and [protocol ADR 0009](https://github.com/open-feature/protocol/blob/main/service/adrs/0009-local-storage-for-static-context-providers.md).

### Notes

- Non-breaking: `domain` is an optional second parameter to `initialize`
- Non-domain-scoped providers retain existing behavior (single instance can back multiple domains; `initialize` runs once)
- Legacy providers that only implement `initialize(context)` remain compatible; the registry inspects the `initialize` signature and omits `domain` when it is not accepted

### Related Issues

Closes #615

Relates to: [open-feature/spec#393](https://github.com/open-feature/spec/pull/393)

## Test plan

- [x] `pytest` (180 tests)

### DIFF
--- a/openfeature/provider/__init__.py
+++ b/openfeature/provider/__init__.py
@@ -35,7 +35,11 @@ class FeatureProvider(typing.Protocol):  # pragma: no cover
 
     def detach(self) -> None: ...
 
-    def initialize(self, evaluation_context: EvaluationContext) -> None: ...
+    def initialize(
+        self,
+        evaluation_context: EvaluationContext,
+        domain: str | None = None,
+    ) -> None: ...
 
     def shutdown(self) -> None: ...
 
@@ -140,8 +144,16 @@ class AbstractProvider(FeatureProvider):
         if hasattr(self, "_on_emit"):
             del self._on_emit
 
-    def initialize(self, evaluation_context: EvaluationContext) -> None:
+    def initialize(
+        self,
+        evaluation_context: EvaluationContext,
+        domain: str | None = None,
+    ) -> None:
         pass
+
+    @property
+    def domain_scoped(self) -> bool:
+        return False
 
     def shutdown(self) -> None:
         pass
--- a/openfeature/provider/_registry.py
+++ b/openfeature/provider/_registry.py
@@ -77,14 +77,10 @@ class ProviderRegistry:
         with self._lock:
             self._reject_domain_scoped_rebind(provider, domain)
             old_provider = self._providers.get(domain)
-            was_bound_elsewhere = provider is self._default_provider or any(
+            already_bound = provider is self._default_provider or any(
                 p is provider for d, p in self._providers.items() if d != domain
             )
-            was_bound_here = (
-                domain in self._providers and self._providers[domain] is provider
-            )
             self._providers[domain] = provider
-            already_bound = was_bound_elsewhere or was_bound_here
             if not already_bound:
                 needs_init = True
                 self._provider_status[provider] = ProviderStatus.NOT_READY
--- a/openfeature/provider/_registry.py
+++ b/openfeature/provider/_registry.py
@@ -7,8 +7,30 @@ from openfeature.event import (
     ProviderEventDetails,
 )
 from openfeature.exception import ErrorCode, GeneralError, OpenFeatureError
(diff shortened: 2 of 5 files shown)


openfeature/provider/__init__.py
```diff
@@ -140,9 +144,17 @@ def detach(self) -> None:
         if hasattr(self, "_on_emit"):
             del self._on_emit
 
-    def initialize(self, evaluation_context: EvaluationContext) -> None:
+    def initialize(
+        self,
+        evaluation_context: EvaluationContext,
+        domain: str | None = None,
```
gruebel, line 150: the problem is this change is not properly backwards compatible, because it fail type-checking with a failed `override` message. my suggestion make it a breaking change, then you don't need to do the `inspect.signature` shenanigans int the registry.

openfeature/provider/_registry.py
```diff
@@ -104,8 +142,32 @@ def shutdown(self) -> None:
     def _get_evaluation_context(self) -> EvaluationContext:
         return get_evaluation_context()
 
+    def _provider_bindings(self, provider: FeatureProvider) -> list[str | None]:
+        bindings: list[str | None] = []
+        if provider is self._default_provider:
+            bindings.append(None)
+        bindings.extend(d for d, p in self._providers.items() if p is provider)
+        return bindings
```
gruebel, line 150: I recommend to do a set comprehension at the beginning with normal provider domains and at the end add the default provider. this would make the setup and afterwards the lookup faster.

openfeature/provider/_registry.py
```diff
@@ -104,8 +142,32 @@ def shutdown(self) -> None:
...
+                )
+            )
+
     def _initialize_provider(
-        self, provider: FeatureProvider, wait_for_init: bool
+        self,
+        provider: FeatureProvider,
+        *,
+        domain: str | None,
+        wait_for_init: bool,
```
gruebel, line 170: why make these keyword only parameters?

openfeature/provider/_registry.py
```diff
@@ -115,22 +177,26 @@ def _initialize_provider(
...
         )
         thread.start()
 
     def _run_initialize(
-        self, provider: FeatureProvider, raise_on_error: bool = False
+        self,
+        provider: FeatureProvider,
+        *,
+        domain: str | None,
+        raise_on_error: bool = False,
```
gruebel, line 196: same here

gruebel (commented):
nice job 🍻 I think we can treat this as a breaking change and therefore make the implementation easier. we have already a breaking change in the pipeline, so no harm done 🙂

PR #621, opened 2026-08-30
fix: reject boolean values for integer flags

### LINKED ISSUE #619: A boolean flag satisfies an Integer request: bool is a subclass of int
## Summary

A boolean flag satisfies an **Integer** request. Evaluating a boolean flag through `get_integer_details` returns `True` with reason `STATIC` and **no error code**, where the specification requires the code default and `TYPE_MISMATCH`.

The application sees a plausible value — `True` behaves as `1` in arithmetic — and no indication that anything went wrong. That is the worst failure mode a feature flag has.

## Cause

`_typecheck_flag_value` in `openfeature/client.py`:

```python
type_map: TypeMap = {
    FlagType.BOOLEAN: bool,
    FlagType.STRING: str,
    FlagType.OBJECT: (dict, list),
    FlagType.FLOAT: float,
    FlagType.INTEGER: int,
}
py_type = type_map.get(flag_type)
...
if not isinstance(value, py_type):
    return TypeMismatchError(f"Expected type {py_type} but got {type(value)}")
```

`bool` is a subclass of `int` in Python, so `isinstance(True, int)` is `True` and the check passes:

```pycon
>>> isinstance(True, int)
True
```

The reverse direction is fine — `isinstance(1, float)` is `False`, so an integer requested as a float is correctly a mismatch — which is why this is easy to miss.

## Reproducing

```python
from openfeature import api
from openfeature.provider.in_memory_provider import InMemoryFlag, InMemoryProvider

api.set_provider(InMemoryProvider({
    "boolean-flag": InMemoryFlag(default_variant="on", variants={"on": True, "off": False}),
}))

details = api.get_client().get_integer_details("boolean-flag", 1)
print(details.value, details.reason, details.error_code)
# True Reason.STATIC None
# expected: 1 Reason.ERROR ErrorCode.TYPE_MISMATCH
```

## Suggested fix

Reject `bool` explicitly when an integer or float was requested:

```python
if flag_type in (FlagType.INTEGER, FlagType.FLOAT) and isinstance(value, bool):
    return TypeMismatchError(f"Expected type {py_type} but got {type(value)}")
```

Worth checking the same place for `FlagType.OBJECT`: `(dict, list)` accepts a `list`, which is correct per the spec's structure type, but is worth a test either way.

## How it surfaced

Building the Python implementation of the cross-language provider conformance suite ([spec#417](https://github.com/open-feature/spec/issues/417)). The suite runs an identical type-mismatch matrix against every provider in every language; this row passes everywhere else and fails only in Python, because only Python has `bool` as a subclass of `int`.

That is a reasonable advertisement for having more than one language implementation: a Java or Go suite could never have caught this.

The suite currently marks the case `xfail(strict=True)` with a pointer to this issue, so it stays visible and un-hides itself automatically once this is fixed.

Hexecu:
## This PR

I reproduced the behavior reported by @aepfli in #619: requesting a boolean flag through `get_integer_details` returns the boolean as a successful evaluation, instead of returning the integer default with `TYPE_MISMATCH`.

The caller gets neither the expected type nor an indication that the flag is misconfigured for that request. The fallback is silently bypassed because Python considers `bool` a subclass of `int`, so `isinstance(True, int)` passes the client's type check.

### Reproducing the problem

This example uses the built-in provider; no flag server or credentials are needed. Run it with `uv run --frozen python repro.py` after saving it as `repro.py` in the checkout:

```python
import asyncio

from openfeature import api
from openfeature.provider.in_memory_provider import InMemoryFlag, InMemoryProvider


async def main():
    api.set_provider_and_wait(
        InMemoryProvider({"flag": InMemoryFlag("on", {"on": True})})
    )
    try:
        client = api.get_client()
        for mode, details in (
            ("sync", client.get_integer_details("flag", 1)),
            ("async", await client.get_integer_details_async("flag", 1)),
        ):
            error = details.error_code.value if details.error_code else None
            print(
                f"{mode}: {details.value!r} ({type(details.value).__name__}), "
                f"{details.reason.value}, {error}"
            )
    finally:
        api.shutdown()


asyncio.run(main())
```

On the base commit, `371aca1`:

```text
sync: True (bool), STATIC, None
async: True (bool), STATIC, None
```

With this patch, the same example returns the supplied default and reports the mismatch:

```text
sync: 1 (int), ERROR, TYPE_MISMATCH
async: 1 (int), ERROR, TYPE_MISMATCH
```

I also checked `False` with a default of `0`. The type matters here: a test that only checks equality would pass even before the fix, since `True == 1` and `False == 0`.

### Why there are two changes

The shared type check now excludes booleans when the requested type is INTEGER. I kept `isinstance` for everything else so this doesn't also reject valid integer subclasses or change how other flag types are handled.

That change alone wasn't enough for the async API. While checking it, I found that the async type-mismatch branch still constructed its result with `resolution.value`. It reported an error, but returned the rejected value anyway. The second change uses `default_value` in that branch, matching the sync behavior.

This also corrects the async fallback for other client-detected type mismatches, which is why the tests cover more than bool-to-int. It doesn't change provider implementations or introduce type coercion. Boolean flags and objects containing booleans continue to work as before.

### Related Issues

Fixes #619. Thanks @aepfli for the reproduction and the conformance test that caught this.

### How to test

The added tests exercise both value and details getters through the real `InMemoryProvider`. They check the exact return type as well as the value, error details and hook behavior: a mismatch must run the error hook, skip the after hook, and give the finally hook the fallback value. There are also passing cases for valid flag values and an integer subclass, to check that the stricter bool handling doesn't affect them.

To run just these tests:

```sh
uv run --frozen pytest tests/test_client.py -q \
  -k 'client_returns_default_on_type_mismatch or client_preserves_matching_flag_types or typecheck_flag_value_accepts_integer_subclasses'
```

I ran the same 39 cases against the unmodified base and this patch. Before the fix, 11 failed and 28 passed: the failures were the two sync bool-to-int cases and the nine async mismatch cases. After the fix, all 39 passed.

The full suite passed on Python 3.10 through 3.14, with 211 tests on each version. The repository's Gherkin suite also passed on each version: 21 scenarios / 84 steps, using its pinned spec and in-memory provider. Ruff, mypy, the remaining pre-commit hooks, and the wheel/sdist build passed locally as well.

```sh
UV_FROZEN=1 uv run --frozen poe test-all
UV_FROZEN=1 uv run --frozen poe e2e
UV_FROZEN=1 uv run --frozen prek run --all-files --show-diff-on-failure
uv build --no-sources
```

### Notes

For a check independent of the tests added here, I ran the in-memory self-test from [python-sdk-contrib#409](https://github.com/open-feature/python-sdk-contrib/pull/409), at `d6de5dc`, against both SDK versions. The case marked as a known failure for #619 becomes `XPASS(strict)` with this patch. Running that unchanged test file with `--runxfail`, so the assertions run normally, gives 24 passed and 5 skipped. The skips are unsupported provider capabilities; the strict marker will need updating when that suite adopts the fixed SDK.

All results above are local macOS runs. I haven't tested against live flagd/OFREP servers, and these results don't replace hosted CI.

### DIFF
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -861,19 +861,6 @@ class OpenFeatureClient:
         default_value: FlagValueType,
         evaluation_context: EvaluationContext | None = None,
     ) -> FlagEvaluationDetails[FlagValueType]:
-        """
-        Resolve a flag asynchronously and validate the returned value.
-
-        Provider-reported errors are returned unchanged. A type mismatch detected
-        by the client returns the caller's default with TYPE_MISMATCH error details.
-
-        :param provider: the provider selected for this evaluation
-        :param flag_type: the requested flag type
-        :param flag_key: the key of the selected flag
-        :param default_value: fallback used for an unknown type or a type mismatch
-        :param evaluation_context: context passed to the provider
-        :return: evaluation details containing the resolved value or fallback
-        """
         get_details_callables_async: Mapping[FlagType, ResolveDetailsCallableAsync] = {
             FlagType.BOOLEAN: provider.resolve_boolean_details_async,
             FlagType.INTEGER: provider.resolve_integer_details_async,
@@ -1004,17 +991,6 @@ class OpenFeatureClient:
 def _typecheck_flag_value(
     value: typing.Any, flag_type: FlagType
 ) -> OpenFeatureError | None:
-    """
-    Check a resolved value against the requested flag type without coercing it.
-
-    Booleans are not integer flag values, even though bool subclasses int in
-    Python. Other subclasses of the expected type remain valid.
-
-    :param value: the value returned by the provider
-    :param flag_type: the requested flag type
-    :return: None for a matching value, TypeMismatchError for an incompatible
-        value, or GeneralError for an unknown flag type
-    """
     type_map: TypeMap = {
         FlagType.BOOLEAN: bool,
         FlagType.STRING: str,
@@ -1025,6 +1001,7 @@ def _typecheck_flag_value(
     py_type = type_map.get(flag_type)
     if not py_type:
         return GeneralError(error_message="Unknown flag type")
+    # bool subclasses int, but booleans are not integer flag values.
     if not isinstance(value, py_type) or (
         flag_type == FlagType.INTEGER and isinstance(value, bool)
     ):
--- a/openfeature/client.py
+++ b/openfeature/client.py
@@ -861,6 +861,19 @@ class OpenFeatureClient:
         default_value: FlagValueType,
         evaluation_context: EvaluationContext | None = None,
     ) -> FlagEvaluationDetails[FlagValueType]:
+        """
+        Resolve a flag asynchronously and validate the returned value.
+
+        Provider-reported errors are returned unchanged. A type mismatch detected
+        by the client returns the caller's default with TYPE_MISMATCH error details.
+
+        :param provider: the provider selected for this evaluation
+        :param flag_type: the requested flag type
+        :param flag_key: the key of the selected flag
+        :param default_value: fallback used for an unknown type or a type mismatch
+        :param evaluation_context: context passed to the provider
+        :return: evaluation details containing the resolved value or fallback
+        """
         get_details_callables_async: Mapping[FlagType, ResolveDetailsCallableAsync] = {
             FlagType.BOOLEAN: provider.resolve_boolean_details_async,
             FlagType.INTEGER: provider.resolve_integer_details_async,
@@ -991,6 +1004,17 @@ class OpenFeatureClient:
 def _typecheck_flag_value(
     value: typing.Any, flag_type: FlagType
 ) -> OpenFeatureError | None:
+    """
+    Check a resolved value against the requested flag type without coercing it.
+
+    Booleans are not integer flag values, even though bool subclasses int in
+    Python. Other subclasses of the expected type remain valid.
+
+    :param value: the value returned by the provider
+    :param flag_type: the requested flag type
+    :return: None for a matching value, TypeMismatchError for an incompatible
+        value, or GeneralError for an unknown flag type
+    """
     type_map: TypeMap = {
         FlagType.BOOLEAN: bool,
         FlagType.STRING: str,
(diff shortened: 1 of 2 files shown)


openfeature/client.py
```diff
@@ -890,7 +903,7 @@ async def _create_provider_evaluation_async(
         if err := _typecheck_flag_value(value=resolution.value, flag_type=flag_type):
             return FlagEvaluationDetails(
                 flag_key=flag_key,
-                value=resolution.value,
+                value=default_value,
```
gruebel: why this change?

openfeature/client.py
```diff
@@ -991,6 +1004,17 @@ def track(
 def _typecheck_flag_value(
     value: typing.Any, flag_type: FlagType
 ) -> OpenFeatureError | None:
+    """
+    Check a resolved value against the requested flag type without coercing it.
+
+    Booleans are not integer flag values, even though bool subclasses int in
+    Python. Other subclasses of the expected type remain valid.
```
gruebel: this info (in short form) should be added to the if condition block.

openfeature/client.py
```diff
@@ -861,6 +861,19 @@ async def _create_provider_evaluation_async(
         default_value: FlagValueType,
         evaluation_context: EvaluationContext | None = None,
     ) -> FlagEvaluationDetails[FlagValueType]:
+        """
+        Resolve a flag asynchronously and validate the returned value.
```
gruebel: also remove any of the doc strings you added. these are internal methods/functions and the text doesn't reflect the whole behavior anyway.

tests/test_client.py
```diff
@@ -424,6 +424,7 @@ async def test_client_type_mismatch_exceptions():
 
 @pytest.mark.asyncio
 async def test_typecheck_flag_value_general_error():
+    """Unknown flag types produce GENERAL rather than TYPE_MISMATCH."""
```
gruebel: no doc strings in test functions, if they don't add substantial information to the test case. the test name should already explain everything needed, if possible.

gruebel (commented):
thanks for the PR, i added a few comments, nothing critical

Hexecu:
Thanks for the review! Removed the added docstrings and moved the bool/int explanation next to the condition. Re-ran the full test suite and Gherkin tests on Python 3.10–3.14; all pass.

gruebel (approved):
thanks 🍻

